PR #9, opened 2020-08-12
docs: fixed BottomSheetView link

remanation:
Just fixed a small issue with the link to BottomSheetView pointing to the flatlist doc instead of the view doc


gorhom:
@remanation thanks for the pr :)

Issue #11, opened 2020-08-12
Web support


marioaks:
Hey!

I just started using this, and so far it's great! Seems like the best bottom drawer out there

# Feature Request

I'm currently trying to use this library with an Expo / React Native Web app. In order to get it working, I used `patch-package` to add a `Touchables.web.js` file and a `ContentWrapper.web.js` file that are exact replicas of their respective ios files. I was wondering if it would be possible to add these two files so patching isn't necessary to support web?

Thanks!

gorhom:
hi @marioaks , if you get it working for web ... well done !! 

please submit a pr and i'll merge it asap 😇

nandorojo:
Hey there, just wondering if you think this will get merged? Thanks for the awesome work on this library, & @marioaks for identifying this issue.

gorhom:
@nandorojo unfortunately i tested this pr, but it is not working properly and i was planning to pick it up but i didn't had the time for it. https://github.com/gorhom/react-native-bottom-sheet/pull/14

nandorojo:
Sounds good, thanks for the heads up!

nandorojo:
Just wondering, is there somewhere you think I should look if I want to try to get this to work on web? Files to change that need work, etc.

gorhom:
hi @nandorojo, to be honest i am not an expert in `react-native-web`. however, you could pick @marioaks pr and i'll try my best to help out :)

Angelk90:
@marioaks : Are there any news?

marioaks:
The PR I submitted was a barebone patch for react-native-web but was never fully functioning. I should have submitted the PR as a draft since I still needed some guidance on how to proceed with the styling of the drawer on web. 

In other words, this PR was never in a mergeable state, but since I never received any sort of feedback on it, I wasn't able to move forward.

Angelk90:
@marioaks : I think you should ask @gorhom.

marioaks:
Hopefully the PR is useful for anyone who tries to get this working for RNW, but I no longer have the time to work on this. Sorry @Angelk90

planecore:
Any updates on this? @gorhom

zhigang1992:
@planecore this actually now works out of the box (almost) see #357 for more detail

nandorojo:
How do you guys handle the UX on a large screen? I wonder if it should become a modal instead of a sheet on a large desktop. Or, do you just set a max width?

gorhom:
[...]

PR #14, opened 2020-08-13
feat: added initial web support and a mostly working web example

marioaks:
Added web fallback files for ContentWrapper and Touchables. Also set up a webpack-dev-server in the
example directory to run on web.

This is all mostly working, but there are a few caveats: 
- Seems there are some styling issues on web. The height of the bottom-sheet seems wrong and the the snapPoints aren't working exactly as they should be because of it.
- MapView seems to not be working for React Native Web so I disabled the map example button.
- Bug in the Animated Shadow example: `Animated.set first argument should be of animatedNode`. Didn't get around to fixing this so I also disabled this button too.

This should be enough of a start though. Maybe you can give me some guidance on what it's supposed to look like so we can get the styling to work on web. If the drawer height is always calculated using the window height and inset, it's very possible that simply wrapping the Bottom Sheet in a <Portal> element would fix the styling problems.

re #11

## Motivation
To use this library with React Native Web

### DIFF
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -1,6 +1,12 @@
 import 'react-native-gesture-handler';
-import { AppRegistry } from 'react-native';
+import { AppRegistry, Platform } from 'react-native';
 import App from './src/App';
 import { name as appName } from './app.json';
 
 AppRegistry.registerComponent(appName, () => App);
+
+if (Platform.OS === 'web') {
+  AppRegistry.runApplication(appName, {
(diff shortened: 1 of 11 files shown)


marioaks:
[This](https://github.com/software-mansion/react-native-reanimated/pull/1088) should fix the Animation bug in the shadow example

ghost:
Waiting for this to be merged :D

PR #30, opened 2020-09-10
fix: handle unstable manualSnapToPoint variable when re-rendering

### LINKED ISSUE #29: Updating state causes methods on bottom sheet to stop working
# Bug

I have some state updates I need to do in my component when I want my bottom sheet to be opened, however when I do this my call to `bottomSheetRef.current?.snapTo(1)` no longer works. If I comment out the state update then it starts working again. The order in which the state update and calling the snapTo method makes no difference either. I have also tried using `bottomSheetRef.current?.expand()` and it has the same issue.

## Environment info

| @gorhom/bottom-sheet         | 1.2.0   |
| react-native                    | 0.63.2   |
| react-native-reanimated         | 1.13.0   |

## Steps To Reproduce

1. Add a function which updates some state on the component and then call one of the methods on the bottom sheet ref
2. Trigger the function
3. State is updated but the method on the bottom sheet is not called
4. Comment out the state update and trigger the function 
5. The method on the bottom sheet now behaves as expected

1. The method on the bottom sheet should be called regardless of component state update

## Reproducible sample code

```
import BottomSheet from '@gorhom/bottom-sheet';
import React, {useMemo, useRef, useState} from 'react';
import {Button, StyleSheet, Text, View} from 'react-native';

const App = () => {
  const [counter, setCounter] = useState(0);
  const bottomSheetRef = useRef<BottomSheet>(null);
  const snapPoints = useMemo(() => [0, '90%'], []);

  const openBottomSheet = () => {
    // comment out setCounter and the bottom sheet opens as expected
    setCounter((prev) => prev + 1);
    bottomSheetRef.current?.snapTo(1);
  };

  return (
    <View style={styles.container}>
      <Button onPress={openBottomSheet} title="Open bottom sheet" />
      <BottomSheet ref={bottomSheetRef} snapPoints={snapPoints}>
        <Text>Hello</Text>
      </BottomSheet>
    </View>
  );
};

const styles = StyleSheet.create({
  container: {
    flex: 1,
    padding: 24,
  },
});

export default App;

```

gorhom:
closes (#29)

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -30,7 +30,6 @@ import {
   State,
 } from 'react-native-gesture-handler';
 import {
-  useValue,
   usePanGestureHandler,
   useTapGestureHandler,
   // ReText,
@@ -186,12 +185,15 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       state: tapGestureState,
       gestureHandler: tapGestureHandler,
     } = useTapGestureHandler();
-
-    const autoSnapTo = useValue<number>(-1);
     //#endregion
 
     //#region animation
-    const { position, currentPosition, currentGesture } = useTransition({
+    const {
+      position,
+      manualSnapToPoint,
+      currentPosition,
+      currentGesture,
+    } = useTransition({
       animationDuration,
       animationEasing,
       contentPanGestureState,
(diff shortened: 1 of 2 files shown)


gorhom:
hi @tom751, could you try this PR

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/manual-snapping-when-rerendering
```

tom751:
I tested and the same bug occurred - as soon as I comment out the state update then it works again.

gorhom:
@tom751 , would mind clear the metro bundler cache and try again, because i have it here working.

ezgif-4-257004df46b0

tom751:
Ah my bad, I cleared the cache and it is working now. Great work!

PR #57, opened 2020-11-04
feat: added dynamic snap point based on content height (#32)

### LINKED ISSUE #32: Dynamic Bottom Sheet height based on Children height
# Feature Request

It would be awesome if you add dynamic height feature based on the children height (and still able to define max height of the bottom sheet). Also the snapPoints prop should be relative to the total Bottom Sheet height, not the screen height.

## Why it is needed

Sometimes we only need to render only a few items inside the bottom sheet. So there's no need to define the bottom sheet height. It should adjust itself according to the children height. And the snapPoints should be relative to the height of the bottom sheet, not the screen height like in the example.

gorhom:
closes (#32) 

### Install

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#feature/dynamic-snap-point
```

### Todo

- [ ]  Allow to provide 1 point when `shouldMeasureContentHeight` is `true`.
- [ ]  Add transition to sheet when content height change.
- [ ]  Add documentation for `shouldMeasureContentHeight`.


EricPKerr:
It would be nice if this supported going from `snapPoints = [0, 200];` (with current index/value being 200), and then switch to `snapPoints = [0, 100, 200];` and it keeps the same current position value, but just internally updates the index.

Similarly, if it goes from `[0, 200, 400]` and the current is 400, and then goes to different values like `[0, 500]` it finds the closest new snapPoint and transitions to that.

likern:
I looked at your PR. It uses Reanimated V1. Should **dynamic snap** later be ported to Reanimated V2 in this branch https://github.com/gorhom/react-native-bottom-sheet/tree/feature/rewrite-in-reanimated-v2?

Could you clarify library roadmap?

gorhom:
hi @likern , thanks for pointing this out. it become clearly that i need to set a roadmap for `v2` that also includes the reanimated v2 version.

I look into planning it later and will put on the project board 👍

likern:
> It would be nice if this supported going from `snapPoints = [0, 200];` (with current index/value being 200), and then switch to `snapPoints = [0, 100, 200];` and it keeps the same current position value, but just internally updates the index.
> 
> Similarly, if it goes from `[0, 200, 400]` and the current is 400, and then goes to different values like `[0, 500]` it finds the closest new snapPoint and transitions to that.

I think instead of fixed snapPoints and all complexity related to dynamic things it's better to allow user to provide `worklet function` which will calculate snapPoint dynamically. So there would be no limitation to how much of them and how to convert between different snapPoint sets.

I did similar in my https://github.com/likern/react-native-enhanced-popup-menu, where the final position on popup is determined at runtime, at moment when user clicked to show popup menu.

User could provide `computeOffset(left, top, width, height)` where in callback he gets coordinated of base component. And he can do any dynamic calculations to do anything with final position.

It think we could do the same. And final snapPoint will be determined dynamically at time when bottomsheet is released.

PR #53, opened 2020-11-02
docs: fixed navigator example link

grifotv:
Simple docs url fix.

PR #59, opened 2020-11-06
Navigator Integration Example

fcandi:
Link not working.

## Motivation
But it would be very interesting to have an example. Could you please provide one?


gorhom:
thanks @fcandi for submitting this pr, however this been fixed with #53

PR #60, opened 2020-11-08
feat: Add possibility to disable tap handler

mkraina:
## Motivation
While quite impressed by the performance of this module, I've added it to my project just to discover problem with react-native touchables (especially on android)
this PR adds possibility to disable the ContentWrapper TapGestureHandler in order to make the react-native touchables work

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -85,6 +85,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       // components
       handleComponent: HandleComponent = Handle,
       backgroundComponent: BackgroundComponent = null,
+      disableTapGestureHandler = false,
       children,
     },
     ref
@@ -397,6 +398,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
           ref={rootTapGestureRef}
           initialMaxDeltaY={snapPoints[Math.max(initialSnapIndex, 0)]}
           style={containerStyle}
+          disableTapGestureHandler={disableTapGestureHandler}
           {...tapGestureHandler}
         >
           <Animated.View style={contentContainerStyle}>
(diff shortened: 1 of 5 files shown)


Ubax:
In our project we have defined design system with many components and due to this change would be able to use this library. @gorhom Would you be able to merge it into version 2?

kazigais:
I would like this feature merged as well

kazigais:
For anyone who needs a workaround - https://gorhom.github.io/react-native-bottom-sheet/troubleshooting#pressables--touchables-are-not-working-on-android

mkraina:
@kazigais 
> For anyone who needs a workaround - https://gorhom.github.io/react-native-bottom-sheet/troubleshooting#pressables--touchables-are-not-working-on-android

This workaround is not applicable when adding this bottom-sheet to large project as for android, they're imported from `react-native-gesture-handler` while on ios `react-native` touchables are used. This messes up the styling as `react-native-gesture-handler` differs between `style` and `containerStyle`. It's quite hard to rewrite half of your codebase just so you can use this lib, hence this PR
@gorhom any feedback on whether this will be merged (if not, why?) 
Thanks in advance :)

gorhom:
@nasmuris first, thanks for submitting this pr.

## The problem 

the problem is disabling `TapGestureHandler` will result in breaking the sheet functionality, and this is something i don't think you would want.

https://user-images.githubusercontent.com/4061838/110862211-210c1c80-82b7-11eb-9e5d-45d6aeccbbad.mov

i am working on a major refactor for `v2` which should fix the teachability issue but it is still in progress and i have been busy getting keyboard handling for `v3` 🤷‍♂️

## Workaround 

although [Pressables / Touchables are not working on Android](https://gorhom.github.io/react-native-bottom-sheet/troubleshooting#pressables--touchables-are-not-working-on-android) this may works for some, but it won't fit other use cases.

but you can use this:

```tsx
<NativeViewGestureHandler disallowInterruption={true}>
   { YOUR TOUCHABLE COMPONENT }
</NativeViewGestureHandler>
```

mkraina:
@gorhom thanks for explanation, I'm using this lib in my project since I've created this PR, but all my contents are horizontal lists so I haven't encountered such an issue you've recorded :)

PR #63, opened 2020-11-12
Only set initialMaxDeltaY on creation of ref to prevent locking of scrollable on re-render

### LINKED ISSUE #51: focusHook causes list to become unscrollable
# Bug

There is an issue when using the `focusHook` prop on a BottomSheetFlatList. Opening the bottom sheet works fine and the list is scrollable, however when the BottomSheet component re-renders it locks the scrolling of the list.

Some investigation reveals that the re-rendering of `BottomSheet` causes `ContentWrapper` to also re-render, re-setting `maxDeltaY`.
I can see that without `focusHook`, the function `refreshUIElements` is usually called which sets the `maxDeltaY` to 0 and enables scrolling. For some reason this is not happening with focusHook.

## Environment info

Both iOS and Android

| @gorhom/bottom-sheet         | 1.4.1   |
| react-native                    | 0.63.3   |
| react-native-reanimated         | 2.0.0-alpha.7   |

## Steps To Reproduce

1. Load the example
2. Open the bottom sheet and scroll the list
3. Tap on a list item which triggers a re-render of the parent component.  
4. Attempt to scroll the list

## Reproducible sample code

```
import React from 'react';
import BottomSheet, { BottomSheetFlatList, TouchableOpacity } from '@gorhom/bottom-sheet';
import { useFocusEffect, NavigationContainer } from '@react-navigation/native';

import { View, AppRegistry, LogBox } from 'react-native';
import { useState } from 'react';
import { SafeAreaProvider } from 'react-native-safe-area-context';
import { createStackNavigator } from '@react-navigation/stack';

const Stack = createStackNavigator();

function Test() {
    const [, reRender] = useState(-1);

    return (
        <SafeAreaProvider>
            <BottomSheet snapPoints={[100, 600]}>
                <NavigationContainer>
                    <Stack.Navigator>
                        <Stack.Screen name="List">
                            {(props) => (
                                <BottomSheetFlatList
                                    focusHook={useFocusEffect}
                                    data={[...new Array(20)]}
                                    keyExtractor={(_, i) => i.toString()}
                                    renderItem={({ index }) => (
                                        <TouchableOpacity onPress={() => reRender(index)}>
                                            <View
                                                style={{
                                                    backgroundColor: 'red',
                                                    height: 50,
                                                    marginTop: 10
                                                }}
                                            />
                                        </TouchableOpacity>
                                    )}
                                />
                            )}
                        </Stack.Screen>
                    </Stack.Navigator>
                </NavigationContainer>
            </BottomSheet>
        </SafeAreaProvider>
    );
}

AppRegistry.registerComponent('Test', () => Test);

```

Thanks!

tankers746:
Fixes https://github.com/gorhom/react-native-bottom-sheet/issues/51

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -395,7 +395,9 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       <>
         <ContentWrapper
           ref={rootTapGestureRef}
-          initialMaxDeltaY={snapPoints[Math.max(initialSnapIndex, 0)]}
+          initialMaxDeltaY={
+            snapPoints[Math.max(currentPositionIndexRef.current, 0)]
+          }
           style={containerStyle}
           {...tapGestureHandler}
         >
--- a/src/components/contentWrapper/ContentWrapper.android.tsx
+++ b/src/components/contentWrapper/ContentWrapper.android.tsx
@@ -13,17 +13,9 @@ const ContentWrapperComponent = forwardRef<
   ) => {
     return (
       <TapGestureHandler
-        ref={(tapHandler) => {
-          tapHandler?.setNativeProps({
-            maxDeltaY: initialMaxDeltaY,
-          });
-          if (typeof ref === 'function') {
-            ref(tapHandler);
-          } else if (ref) {
-            ref.current = tapHandler;
-          }
-        }}
+        ref={ref}
         maxDurationMs={1000000}
+        maxDeltaY={initialMaxDeltaY}
         shouldCancelWhenOutside={false}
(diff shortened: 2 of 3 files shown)


gorhom:
hi @tankers746 , thanks for submitting this pr :)

after investigating this issue, i notice `initialMaxDeltaY` is using `initialSnapIndex` which is static value that user set as a prop.

and to fix it, i would suggested change

```tsx
          initialMaxDeltaY={snapPoints[Math.max(initialSnapIndex, 0)]}
```

to 

```tsx
          initialMaxDeltaY={
            snapPoints[Math.max(currentPositionIndexRef.current, 0)]
          }
```

tankers746:
Great, i've updated that

tankers746:
Not needed in alpha 2

PR #70, opened 2020-11-19
fix: return flipper back to make example working

likern:
## Motivation

I tried to build and run example application from `rewrite-in-reanimated-v2` branch.
It failed because of commented code in PR.
After commenting out example starts working again.


gorhom:
thanks @likern for submitting this pr 👍

PR #80, opened 2020-11-21
chore(lint-staged): install and configure lint-staged

likern:
## Motivation

`lint-staged` allows running linting only on staged files. 
This is especially useful when lint-staged is run from `husky:pre-commit` hook, because husky runs linting on all files, even those which are not part of commit.


gorhom:
thanks @likern for submitting this pr 👍

PR #81, opened 2020-11-22
feat: allow dynamic snap points

### LINKED ISSUE #32: Dynamic Bottom Sheet height based on Children height

### LINKED ISSUE #77: snapPoints it differs a lot between operating systems
# Bug

90%
Schermata 2020-11-20 alle 22 24 05

100%

Schermata 2020-11-20 alle 22 30 53

Hi @gorhom , 

I am using your component on different devices, I tried to put snapPoints = "90%".
I noticed that it differs a lot between the two, as can be seen from the image.

I tried with higher values (100%), it seems that on android it works fine but on ios it gives problems, it goes further.

## Environment info

| @gorhom/bottom-sheet         | 1.4.1   |
| react-native                    | 0.63.3 |

## Reproducible sample code

```
{
          snapPoints: ['90%'],
          animationDuration: 10,
          overlayComponent: BottomSheetOverlay,
          overlayOpacity: 0.5,
          dismissOnOverlayPress: true,
        }
```

gorhom:
## Motivation

this pr will allow `snapPoints` to be more reactive to updates and improve snap points calculating position.

### Test
```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#allow-dynamic-snap-points

```


likern:
@gorhom How to test it? Can't understand how to use this feature. Is it reflected in example App?

likern:
Is dynamic snap point is what located at `src/screens/static/BasicExample.tsx`?
I see it's not shown in example application.

likern:
I have feeling that bottom-sheet handle is not synced with content, expecially on shaking.

https://drive.google.com/file/d/1DZkq2LoF-uaRpzaga475wzysMVlj2MvI/view?usp=sharing
Sorry for the video quality.

gorhom:
hi @likern , thanks for tasting this pr

>I have feeling that bottom-sheet handle is not synced with content, expecially on shaking.

this is a known issue on Android, i will work on a fix later.

likern:
@gorhom  Thank you! I have two more questions.

Will this version work for Reanimated V2? My reanimated version is `2.0.0-alpha.8`. I don't use old `useCode()`-like API, but as I understand V2 should fully support V1 API, so even `react-native-bottom-sheet` V1 should work for `react-native-reanimated-2.0.0-alpha.8`. Am I correct?

And could you give more examples how to use bottom-sheet based on content? For example just show bottom-sheet with one snap point of content height?

gorhom:
@likern this pr for `v2` which is written in Reanimated `v1` compatible with Reanimated `v2`. However, i have put this feature in `v3` roadmap which i'll need to rewrite it in Reanimated `v2`.

>And could you give more examples how to use bottom-sheet based on content? For example just show bottom-sheet with one snap point of content height?

i will add one more example for dynamic snap point later 👍

gorhom:
@likern added dynamic snap point example ,, hope it helps 👍

likern:
@gorhom I think gestures are not interruptable.
If I, during bottom sheet animation (when it transitions from one snap point to another), tap on it (without panning) - it doesn't pause / interrupt it's animation.

gorhom:
@likern i think that is an expected behaviour,, right. but also you could create an issue and we discuss it from there :)

stephenlaughton:
@gorhom this looks sick! one question. do you know of any way to make Modal snap points reactive? the use case is presenting modal confirmation dialogues that have dynamic height based on content. 

really awesome work on this library by the way 🔥 !

gorhom:
@stephenlaughton good point , I'll look into it

PR #139, opened 2021-01-02
feat: added over drag support

gorhom:
## Motivation

inspired by `reanimated-bottom-sheet` over drag feature

https://user-images.githubusercontent.com/4061838/103460626-08601300-4d18-11eb-9edf-38956a831132.mp4

## Changelog

- added `enableOverDrag`  & `overDragResistanceFactor` props
- added `animationConfigs` to customise animation configuration and type.
- added `useBottomSheetSpringConfigs` & `useBottomSheetTimingConfigs` to create a memoized animation configs


TransLover:
It'd be nice if one could over drag it to the upside by dragging anywhere on the modal instead of just the very top, like the way instagram does it

pouyarezvani:
@lg9302 @gorhom This would be very Awesome!
Can we expect this to happen @gorhom?

pouyarezvani:
@gorhom Any chance this [comment above](https://github.com/gorhom/react-native-bottom-sheet/pull/139#issuecomment-859205196) can land in V4?

gorhom:
@lg9302 @pouyarezvani this is impossible - at least for any scrollable - because it will block the scrollable from scrolling 🤷‍♂️

however, you can overdrag to the bottom from any scrollable/content 👏

TransLover:
> @lg9302 @pouyarezvani this is impossible - at least for any scrollable - because it will block the scrollable from scrolling 🤷‍♂️
> 
> however, you can overdrag to the bottom from any scrollable/content 👏

Yes it would be impossible for any scrollable content, but for the regular modal sheet it is possible

pouyarezvani:
@grifotv @gorhom Yeah it makes sense for it to not work in scrollable views. But what about if a view is not scrollable? Can it be achieved there? Could you possibly hint me how you are doing your over-drag which is only for the handle section so maybe I can tweak it to work with the whole body of the modal?

PR #142, opened 2021-01-03
fix: prevent scrollable glitches

gorhom:
## Motivation

Refactored `useScrollableInternal` to prevent scrollable scrolling glitches


osdnk (changes requested):
This commit introduces again regression I fixed with https://github.com/gorhom/react-native-bottom-sheet/pull/141/files

PR #151, opened 2021-01-05
fix: fixed the glitch when snapping animation is running on android

gorhom:
## Motivation

this should fix the glitches that occurs on android caused by `Easing.exp` from reanimated v2


osdnk:
amazing

PR #153, opened 2021-01-06
feat: wrap android's bottom sheet content wrapper with ScrollView

osdnk:
I noticed that wrapping in additional ScrollView (that technically doesn't do anything but activating immediately) makes it properly interacting with outer ViewPager. 

before: 

https://streamable.com/pvle39

after:

https://streamable.com/y8lvjm

### DIFF
--- a/src/components/bottomSheetContentWrapper/BottomSheetContentWrapper.android.tsx
+++ b/src/components/bottomSheetContentWrapper/BottomSheetContentWrapper.android.tsx
@@ -1,4 +1,5 @@
 import React, { forwardRef, memo } from 'react';
+import { ScrollView } from 'react-native';
 import isEqual from 'lodash.isequal';
 import Animated from 'react-native-reanimated';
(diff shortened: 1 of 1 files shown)


src/components/bottomSheetContentWrapper/BottomSheetContentWrapper.android.tsx
```diff
@@ -19,16 +20,23 @@ const ContentWrapperComponent = forwardRef<
...
-    <AnimatedTapGestureHandler
-      ref={ref}
-      maxDurationMs={1000000}
-      shouldCancelWhenOutside={false}
-      onGestureEvent={handleGestureEvent}
-      // @ts-ignore
-      animatedProps={animatedProps}
+    <ScrollView
+      style={{ height: '100%' }}
+      contentContainerStyle={{ height: '100.1%' }}
```
osdnk, line 25: this is enough for activating, but actually has no real noticeable impact

PR #237, opened 2021-01-27
feat: added bottom inset

### LINKED ISSUE #95: Footer Component
Is there a way of implementing a footer component with this bottom sheet? essentially almost behave like the `handleComponent` prop but it allows you to pass in a component thats always present at the bottom of the sheet regardless of the snapPoints?

DawidVanderhoven-TomTom:
I am very new to reanimated so this may be a completely incorrect way of approaching this, however this is my attempt so far. This allows you to create space below the bottom sheet in case of a situation where a footer or a bottom menu may need to be present.

This resolves #95


hanno-ferrannegre:
@DawidVanderhoven-TomTom what about for `v3`? Any ideas?

ferrannp:
@DawidVanderhoven-TomTom what about for v3? Any ideas?

DawidVanderhoven-TomTom:
@ferrannp I am using v2 at the moment but perhaps if you attempt to do something similar on [here](https://github.com/gorhom/react-native-bottom-sheet/blob/v3/src/components/bottomSheet/BottomSheet.tsx#L447) subtracting the amount of space you want below would be worth a shot however this is untested and I am still not totally sure if this is the correct approach.

gorhom:
thanks @DawidVanderhoven-TomTom for submitting this pr, i will review it shortly

DawidVanderhoven-TomTom:
@gorhom I have now also provided an example and added the `safeHandleHeight` to the `handleClose` as I found that I could sometimes see the handle when stacking modals when using it with `bottomInset`.

gorhom:
hi @DawidVanderhoven-TomTom , i have tested you pr , the example that you added works fine , but other modal examples got broken.

i understand the need for this, and i would suggest to change the approach a bit, 

you could pass `topInset` & `bottomInset` to `BottomSheetContainer` and  set them as the root container `top` and `bottom` like :

```tsx
const containerStyle = useMemo(
  () => [
    styles.container,
    {
      top: topInset,
      bottom: bottomInset,
      overflow: 'hidden',
    },
  ],
  [bottomInset, topInset]
);
```

and it should achieve the goal 👏

gorhom:
the solution above could be implemented easily in `v3` too

DawidVanderhoven-TomTom:
Thanks @gorhom!

gorhom:
thanks @DawidVanderhoven-TomTom for submitting this pr <3

hanno-ferrannegre:
@gorhom plans to porting it to v3?

ferrannp:
@gorhom plans to porting it to v3?

chybisov:
@gorhom could you please suggest the way how to achieve same functionality on v3? This is critical feature, because now it's not clear how to address these issues:

1) Integrate with something that needs to be below bottom sheet (e.g. bottom navigation bar).
2) On iPhone X/11/12 we have bottom line and bottom sheet should take into account its height.

I'm not sure where this should be done - on lib side or our own, but tips to this would be so much helpful. Thank you!

PR #239, opened 2021-01-27
fix: stabilise animated callback node variables

### LINKED ISSUE #208: Backdrop flashes on mount
# Bug
Adding a backdrop on v2 seems to causes the initial backdrop to flash on mount even though the modal is closed. You can see here, where once I tap "mount" it flashes the backdrop quickly. If I remove the backdrop, it doesn't show up. I've tried doing a custom, more basic backdrop and it doesn't seem to work either (same thing). Note: it's fairly quick here, but in a larger app it seems to take even longer to disappear

Kapture 2021-01-15 at 08 43 53

I did notice that this issue #118 seems to be related, but I'm on the latest version and still having issues. I tried adding some debugging logs but couldn't get more from it. I have a feeling it's related to the `animatedIndex` value being passed in? That value controls opacity of the backdrop and if it's not 0 on mount it'll flash (what seems to be at full opacity). 

## Environment info

| @gorhom/bottom-sheet         | 2.0.3   |

## Steps To Reproduce
Go to android (IOS doesn't have the problem on the snack simulator, but if you do it on real simulator or real device you'll see it, press mount, you'll see the screen flicker:

https://snack.expo.io/7UoD9PU0Z

On mount, if the modal is closed I.e. index === -1, we should not show the backdrop flash. The animation code is correct, but the mounting logic seems a little odd

## Reproducible sample code
https://snack.expo.io/7UoD9PU0Z

gorhom:
close #208 

## Motivation

with this pr, `animatedIndex` and `animatedPosition` won't be interpolated until the layout is calculated.

## Installation

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/animated-variables-flickering
```

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -262,16 +262,25 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
         adjustedSnapPointsIndexes.push(-1);
       }
 
-      return interpolate(position, {
-        inputRange: adjustedSnapPoints,
-        outputRange: adjustedSnapPointsIndexes,
-        extrapolate: Extrapolate.CLAMP,
-      });
-    }, [position, safeContainerHeight, snapPoints]);
+      return cond(
+        animatedIsLayoutReady,
+        interpolate(position, {
+          inputRange: adjustedSnapPoints,
+          outputRange: adjustedSnapPointsIndexes,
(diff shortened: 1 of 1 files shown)


theohdv:
Works perfectly in my case ! Thanks a lot for your reactivity

gorhom:
thanks for testing it @theohdv

gorhom:
this should be fixed with `v2.0.7` 🎉

PR #244, opened 2021-01-28
fix: prevent modal crash when navigating

### LINKED ISSUE #240: BottomSheetModalProvider crashing in some circumstances when dismissing a modal
# Bug

`null is not an object (evaluating 'sheetsQueueRef.current[sheetsQueueRef.current.length - 1].ref.current.restore')` when calling `bottomSheetModalRef.dismiss()`.
It happens when the modal is somehow unmounted before the animation ends (in my case because the page is unmounted from the stack of react-navigation), so 

## Environment info

| @gorhom/bottom-sheet         | 2.0.5   |
| react-native                    | 0.63.4  |

## Steps To Reproduce

To reproduce the bug you need to unmount the page containing the modal before the animation ends, so that ModalProvider will use a null ref.
I tried to add a setTimeout of 1 second (so the animation ends) before unmounting the page (before calling `navigation.goBack()`) and everything works fine, so I think that the modal gets removed from sheetsQueue only when animation ends.

I tested with ref.dismiss() and dismiss and dismissAll from `useSheetModal()` and it always crashes in the same way:
1. open the modal
2. dismiss the modal and unmount the page containing it (navigation.goBack) before animation ends
3. open another modal
4. when trying to dismiss it (even with drag down) I get the error because bottomSheetModalProvider tries to restore the modal which was unmounted in step 2

## Reproducible sample code

I can't provide it right now but I will add it later if you really need it

gorhom:
close #240 

## Motivation

to provide better handling when unmounting due to navigation change.

https://user-images.githubusercontent.com/4061838/106208485-c1a0f400-61c3-11eb-9265-46e299501578.mp4

## Installation 

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/prevent-modal-crash-when-navigating
```

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -9,7 +9,7 @@
     "start": "react-native start"
   },
   "dependencies": {
-    "@gorhom/portal": "^0.1.4",
+    "@gorhom/portal": "^0.2.0",
     "@gorhom/showcase-template": "^1.0.2",
     "@react-native-community/blur": "^3.6.0",
     "@react-native-community/masked-view": "0.1.10",
--- a/example/yarn.lock
+++ b/example/yarn.lock
@@ -818,10 +818,10 @@
   dependencies:
     "@types/hammerjs" "^2.0.36"
 
-"@gorhom/portal@^0.1.4":
-  version "0.1.4"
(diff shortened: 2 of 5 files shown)


gorhom:
@alessiocancian could you please test this pr :)

alessiocancian:
@gorhom I'm testing it and it seems to be fixed

PR #253, opened 2021-02-01
fix: prevent crash when resizing container height on Android

### LINKED ISSUE #251: Animate on mount causing hard crash when container height changes on Android
# Bug

There is an issue when using the animateOnMount prop on Android. Changes to the container height cause the app to crash with a SIGABRT Abort Program. 

## Environment info

| @gorhom/bottom-sheet         | 3.2.1   |
| react-native-reanimated         | 2.0.0-rc.2   |

## Steps To Reproduce

1. Run the example
2. Tap the "Decrease container height" button
3. App crashes

## Reproducible sample code

```
import BottomSheet from '@gorhom/bottom-sheet';
import React, { useState } from 'react';
import { AppRegistry, TouchableOpacity, Text, Dimensions } from 'react-native';

const App = function App() {
    const [containerHeight, setContainerHeight] = useState(Dimensions.get('window').height);
    return (
        <BottomSheet
            index={0}
            snapPoints={[100, 500]}
            animateOnMount
            containerHeight={containerHeight}>
            <TouchableOpacity onPress={() => setContainerHeight(height => height - 100)}>
                <Text>Decrease container height</Text>
            </TouchableOpacity>
        </BottomSheet>
    );
};

AppRegistry.registerComponent('App', () => App);

```

Cheers,
Tom

gorhom:
close #251 

## Motivation

for some reason when when passing a parameter from an array, in android it will crash

```ts
// CRASH
requestAnimationFrame(() => runOnUI(animateToPoint)(snapPoints[currentIndexRef.current]));

// FIX
const newSnapPoint = snapPoints[currentIndexRef.current];
requestAnimationFrame(() => runOnUI(animateToPoint)(newSnapPoint));

``` 

## Installation 

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/prevent-crash-on-container-height-resizing
```

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -366,7 +366,8 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
             snapPoints.length - 1
           }`
         );
-        runOnUI(animateToPoint)(snapPoints[index]);
+        const newSnapPoint = snapPoints[index];
+        runOnUI(animateToPoint)(newSnapPoint);
       },
       [animateToPoint, snapPoints]
     );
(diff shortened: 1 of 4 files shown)


gorhom:
hi @tankers746, could you please test this pr :)

tankers746:
Thanks thats fixed the issue
However the style change to `BottomSheetContainer.tsx` is now causing the sheet to be cut off. This is however only happening on my device with a notch.
Screenshot_20210202-142911_VROMO Driver

gorhom:
the container will always align to the top, and that what will happened when you override its height

tankers746:
This is a problem with the automatic container height.

However I can see now this is just a case of RN not playing nice with my translucent status bar, for some reason the onLayout height in BottomSheetContainer is excluding the height of the status bar.

The change above does change the existing behaviour but I can work around it.

gorhom:
if you provide a container height it will no longer do automatic layout measuring

PR #257, opened 2021-02-01
fix: expand sheet when we closed it many times

benhartouz:
Can't expand the sheet when we closed it many times

## Motivation

https://user-images.githubusercontent.com/15239702/106521144-156e4e80-64de-11eb-9d79-e605ac549b81.mov


gorhom:
thanks @benhartouz for submitting this pr

PR #309, opened 2021-02-25
Fix type export of internal modal

thgh:
## Motivation

Using expo-next triggers this error:
```
error - ./node_modules/@gorhom/bottom-sheet/lib/module/contexts/index.js
Attempted import error: 'BottomSheetModalInternalContextType' is not exported from './modal/internal'.
```


gorhom:
thanks @thgh for submitting this pr

PR #337, opened 2021-03-10
feat: adds optional snapPointIndex parameter to modal present method

krakz999:
Adds an optional parameter to the modal's `present` method to be able to present the modal on the desired index, rather than the one defined as prop for the `BottomSheetModal` component.

Also fixes an issue with stacked modals where the modals could stuck in a closing state when rapidly switching between them, or "grabbing" them in-between their animation.

## Motivation

Wanted to present the modal on a different snap point depending on the scenario. It was counter intuitive to use the prop and maintain a state to keep track of the indexes. Other issue was modals getting "stuck" in their state when swapping them rapidly, breaking the modal stack.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -363,7 +363,10 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
         );
         if (isClosing.current && !force) {
           return;
+        } else {
+          isClosing.current = false;
         }
+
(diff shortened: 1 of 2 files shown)


krakz999:
The reason for the modals being stuck is because isClosing.current value does not change back to true. 

`handleRestore` in BottomSheetModal seems to restore the modal by invoking `snapTo` on it with `forced = true`, so I decided to reset the `isClosing` value in the `snapTo` method when it is forced.

If this feels dirty, maybe I can expose a restore method directly from BottomSheet that does the same, instead of having it as a side effect @gorhom ?

PR #350, opened 2021-03-17
feat: changes all arrays to ReadonlyArray

Ubax:
## Motivation

When using arrays in props it is often a better solution to use `ReadonlyArray`. This way, the array cannot be modified in the component that uses it and it is easier to find bugs.


gorhom:
thanks @Ubax, for submitting this pr 👏

PR #367, opened 2021-03-25
example with nested flatlist

likern:
## Motivation

This is example to discuss problems I have integrating my https://github.com/breeffy/react-native-calendar with react-native-bottom-sheet.

I was trying to utilize `waitFor` to make react-native-bottom-sheet waiting nested flatlist, but it doesn't work for me, if my code is correct.

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -144,6 +144,16 @@ function App() {
           />
 
           <Stack.Screen
+            name="Advanced/NestedScrollableExample"
+            options={{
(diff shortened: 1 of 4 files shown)


princefishthrower:
@likern - Not only am I having problems integrating the two, I get a full crash whenever I put a `<Calendar/>` in the `<BottomSheet/>` did you ever find a solution here?

EDIT: Ah, I see the referenced repo is not the standard `react-native-calendar` repo...

PR #378, opened 2021-03-29
fix: backdrop random appearance

### LINKED ISSUE #375: [v3] Backdrop not always appearing
The backdrop isn't always appearing, I tested this with the example app too (I disabled the alert to make it easier to see the difference)

https://user-images.githubusercontent.com/1611055/112805394-d2f86680-9075-11eb-8d56-98b7ce44894d.mov

gorhom:
closes #375 

## Motivation

For some reason `transform` is not stable when component mount then animate

## Installation

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/backdrop-appearance
```

### DIFF
--- a/example/src/screens/modal/BackdropExample.tsx
+++ b/example/src/screens/modal/BackdropExample.tsx
@@ -73,7 +73,7 @@ const BackdropExample = () => {
         onDismiss={handleDismiss}
         backdropComponent={renderBackdrop}
       >
-        <ContactListContainer title="Modal FlatList" type="FlatList" />
+        <ContactListContainer title="Modal View" type="View" />
       </BottomSheetModal>
     </View>
   );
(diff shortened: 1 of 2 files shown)


gorhom:
@Johan-dutoit could you try this pr :)

Johan-dutoit:
Testing this in the bottom sheet example project resulted in the expected behaviour, so LGTM.

I do need to see what's happening on my local project though, as it removes all my pointer events (on the contents below, from the initial render)... could just be a bad implementation on my part.

Johan-dutoit (approved):

gorhom:
@Johan-dutoit i think it is better to investigate your usecase, is it possible to provide a sample reproducible snack ?

Johan-dutoit:
After further investigation, it was an old bottom sheet that wasn't implemented correctly.

Issue #364, opened 2021-03-24
Is BottomSheet FlatList still using getNode()


MagnusEklof:
## Ask your Question

I am using the latest version of flatlist but with newer version of react-native and I get a warning that "FlatList: Calling getNode() on the ref of an Animated component is no longer necessary. You can now directly use the ref instead. This method will be removed in a future release."

Can you confirm? And if so can you fix this?

RealDyllon:
I'm getting the same warning as well, on React Native 0.64.0

oneji:
Can u please accept the PR to solve this warning? Or as I can see this repo isn't maintaned anymore... @vonovak @richkeenan @thgh @olliebrennan @grifotv

rrebase:
@oneji I highly recommend https://github.com/ds300/patch-package for these types of quick fixes.

This repo is definitely in active development and @gorhom is doing a fantastic job 👏. Considering the popularity of the library and amount of use cases it just takes time to address all of them while building new features.

Issue #363, opened 2021-03-24
Is getNode() still being used in the BottomSheetFlatList


MagnusEklof:
I am using the version of flatlist but with newer version of react-native and I get a warning that "FlatList: Calling getNode() on the ref of an Animated component is no longer necessary. You can now directly use the ref instead. This method will be removed in a future release."

Can you confirm? And if so can you fix this?

PR #412, opened 2021-04-21
chore: remove calls to getNode()

rrebase:
This PR removes calls to `getNode()` in `v2`

## Motivation

Resolve the warning `"Calling getNode() on the ref of an Animated component is no longer necessary.`

Related: https://github.com/gorhom/react-native-bottom-sheet/issues/364, https://github.com/gorhom/react-native-bottom-sheet/issues/363

Seems to be resolved in `v3` with https://github.com/gorhom/react-native-bottom-sheet/pull/166

### DIFF
--- a/src/components/flatList/FlatList.tsx
+++ b/src/components/flatList/FlatList.tsx
@@ -52,7 +52,7 @@ const BottomSheetFlatListComponent = forwardRef(
 
     // effects
     // @ts-ignore
-    useImperativeHandle(ref, () => scrollableRef.current!.getNode());
+    useImperativeHandle(ref, () => scrollableRef.current);
     useFocusHook(handleSettingScrollable);
 
     // render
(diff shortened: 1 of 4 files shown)


gorhom:
thanks @rrebase for submitting this pr 👏

my only concern is that there may be users that still using < 0.64 , and this may break the library. Would you mind testing this pr on RN v0.63 ?

rrebase:
Yeah, a valid concern. It's deprecated since RN v0.62 https://reactnative.dev/blog/2020/03/26/version-0.62#deprecations

However, I now tried with RN v0.63 and while it worked with reanimated v2, it broke when reanimated v1 was installed. Relevant change: https://github.com/software-mansion/react-native-reanimated/commit/6a6bded5cfc6d9b7e732fd6393758e954b24fcb0

So unless there's an easy fix for both versions (which `v2` supports) it makes sense to just close this and recommend upgrading to `v3` 🤔

PR #430, opened 2021-05-06
refactor: converted all internal state/memoized variables to reanimated shared values.

gorhom:
https://github.com/gorhom/react-native-bottom-sheet/discussions/427

## Motivation

I have refactored the library to use shared values to store and update layout values.

### Notice
Calculating snap points position had change in this branch, now the library will count the handle height when positioning the sheet. This is a breaking change and will require developers to carefully monitor the change.

This is how you can visualise it:

snap-point-calculation

### Installation

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#refactor/convert-variables-to-shared-values
```


tambadiya:
Hello @gorhom stack navigator in bootomsheet and I have different action button to all stack navigator screen so can I use BottomSheetFooter only in stack navigator child screens.

PR #448, opened 2021-05-19
fix: allow passing custom style to backdrop component

hanselabreu:
## Motivation

Allow modification of the style property. This will allow overriding the `backgroundColor`.

This was fixed on #211 but then reverted on #378 (probably by accident).


vicolsson (approved):
Just what I need 👍

gorhom:
@hanselabreu thanks for submitting this pr !

PR #495, opened 2021-06-19
feat: Added bottomInset to BottomSheetBackdropContainer

kickbk:
## Motivation

A newly added feature to BottomSheet is the bottomInset which offsets a BottomSheet from the bottom. However, the backdrop component should also receive the same offset. This PR adds the bottomInset to `BottomSheetBackdropContainer` by setting the absolute bottom in styles to the bottomInset, if it exists, or defaults to 0.
Also removed an unused `useMemo` in `BottomSheetHandleContainer`


kickbk:
This PR seems to have issues. Will redo.

PR #498, opened 2021-06-21
feat: Added bottomInset to BottomSheetBackdropContainer

kickbk:
## Motivation

A newly added feature to BottomSheet is the `bottomInset` which offsets a BottomSheet from the bottom. However, the backdrop component should also receive the same offset. This PR adds the bottomInset to `BottomSheetBackdropContainer` by setting the absolute bottom in styles to the bottomInset, if it exists, or defaults to 0.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -1252,6 +1252,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
           animatedIndex={animatedIndex}
           animatedPosition={animatedPosition}
           backdropComponent={backdropComponent}
+          bottomInset={bottomInset}
         />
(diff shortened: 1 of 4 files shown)


kickbk:
In some cases one might want to offset the sheet but not the backdrop. In that case we could add a boolean prop `offsetBackdrop`, or possibly a `backdropBottomOffset` instead to apply a numerical value instead of assuming they want it or that it should be the same as the BottomSheet bottomInset?

gorhom:
The problem that i see, what if we want to add `topInset` ? then we will need to update the code again. for me i would override it this way :
```tsx
  props => (
    <BottomSheetBackdrop {...props} style={{ bottom: bottomInset }} />
  ),
  [bottomInset]
);
```
where you have the full control of all insets ( top, left, right, bottom )

kickbk:
Yes, that's better

Issue #411, opened 2021-04-20
BottomSheetTextInput that has multiline enabled, does not push bottom sheet up when focused and keyboard shows.


jonesaustindev:
# Bug
BottomSheetTextInput that has multiline enabled, does not push bottom sheet up when focused and keyboard shows. If I remove the ability to have multiline enabled, it works as expected with no issue.

https://user-images.githubusercontent.com/36755092/115439911-b662d080-a1d4-11eb-96e7-8a5fad951076.mov

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet         | "^3.6.3"   |
| react-native                    | "0.63.4"  |
| react-native-reanimated         | "^2.0.0"   |
| react-native-gesture-handler    | "^1.9.0"  |

## Steps To Reproduce
Setting up a simple implementation of the `BottomSheetModal` will show how it behaves.

1. Make one `BottomSheetTextInput` have multiline while the other does not.
2. Tap the one without `multiline` to see it function correctly.
3. Tap the one with `multiline` to see it does not slide up the bottom sheet.

Describe what you expected to happen:

1. Keyboard should push up the bottom sheet when using `BottomSheetTextInput` with `multiline` enabled.

## Reproducible sample code
Note: This is inside a wrapper component that just has data/state management with a `View` with the style `{{ flex: 1, justifyContent: 'center', alignItems: 'center' }}`.
```
import React, { useContext, useMemo, useRef } from 'react';
import { StyleSheet, ScrollView, TouchableOpacity, Platform } from 'react-native';
import { useSafeAreaInsets } from 'react-native-safe-area-context';
import { useNavigation } from '@react-navigation/core';
import {
  BottomSheetModal,
  BottomSheetModalProvider,
  BottomSheetBackdrop,
  BottomSheetTextInput,
  BottomSheetScrollView,
} from '@gorhom/bottom-sheet';

import { Text } from '../components/common/Text';
import { View } from '../components/common/View';
import { Button } from '../components/common/Button';
import { ReorderIcon } from '../assets/svg/ReorderIcon';
import { ACTIVE_OPACITY } from '../constants';
import { localizedStrings } from '../localization/localizedStrings';
import { colors, globalStyles, SCREEN_HEIGHT, SCREEN_WIDTH } from '../styles/globalStyles';
import { GradientFooter } from '../components/common/GradientFooter';
import { HorizontalLine } from '../components/common/HorizontalLine';

[...]

PR #565, opened 2021-08-06
Remove shouldHandleKeyboardEvents

dickfickling:
## Motivation

Keyboard avoidance functionality doesn't work for multiline text inputs (see https://github.com/gorhom/react-native-bottom-sheet/issues/411). For now, my solution is to remove the `shouldHandleKeyboardEvents` gate. I'd love to understand more about why that shared state variable exists, so I can come up with a better solution.

note: this issue exists on v4 as well


gorhom:
thanks @dickfickling for submitting this pr, however `shouldHandleKeyboardEvents` is needed to determine the source of the keyboard appearance event. I have fixed this issue for `v4` [here](https://github.com/gorhom/react-native-bottom-sheet/commit/e21d6762a929c6eaaf64e95d8af2934cc8b3a703) and it will be release on the next release today 👍

PR #571, opened 2021-08-11
feat: add custom accessibility support

### LINKED ISSUE #548: [Proposal] Improving accessibility
## Ask your Question

Hey, thank you for creating such a promising library!
Some context - my team works for a large company and this library has caught our eyes as it has almost everything that we need. Since accessibility is one of our top priorities - we wanted to ask if we could contribute by improving the accessibility of the bottom sheet?

## Proposal

Here is a list of changes we are willing to do. Please let us know if the proposal interests you.

[BottomSheetBackdrop] make the following props overridable:
* accessible;
* accessibilityRole;
* accessibilityLabel;
* accessibilityHint.

[BottomSheet]
* make the following props overridable:
    * accessible;
    * accessibilityRole;
    * accessibilityLabel;
* add the following props:
    * announcePositionChange?: boolean:
        * if disabled - does not announce the position changes on screen reader;
        * by default is enabled (true);
    * onPositionChangeAnnounce?: (positionInScreen) => string;
        * allows the customisation of position change announcement for screen reader;
        * takes a callback that returns a string that will be read out by the screen reader.

[BottomSheetHandleContainer] move accessibility props to [BottomSheetHandle] and make the following props overridable:
* accessible;
* accessibilityRole;
* accessibilityLabel;
* accessibilityHint.

When setting accessibility to false on any of the components - screen reader should not detect the elements.
Fix bug where if modal is dismissed it announces that it has snapped to 10%.

nmuranovas:
## Motivation

This PR adds custom accessibility props to BottomSheet, BottomSheetHandle and BottomSheetBackdrop components without introducing breaking changes.

I posted the proposal changes in - https://github.com/gorhom/react-native-bottom-sheet/issues/548


src/types.d.ts
```diff
@@ -49,3 +54,10 @@ export type ScrollableRef = {
   type: ScrollableType;
 };
 //#endregion
+
+export interface NullableAccessibilityProps extends AccessibilityProps {
```
gorhom, line 58: why not just use `Pick`
```ts
export interface NullableAccessibilityProps
  extends Pick<
    AccessibilityProps,
    | 'accessible'
    | 'accessibilityLabel'
    | 'accessibilityHint'
    | 'accessibilityRole'
  > {}
```

src/components/bottomSheet/types.d.ts
```diff
@@ -140,7 +141,23 @@ export type BottomSheetProps = {
...
   children: (() => React.ReactNode) | React.ReactNode[] | React.ReactNode;
-} & BottomSheetAnimationConfigs;
+
+  // accessibility
+  /**
+   * Enables announcement of the snap position change on screen readers.
+   * @type boolean
+   * @default true
+   */
+  announceChangeForAccessibility?: boolean;
```
gorhom: please rename this to `enableAccessibilityChangeAnnouncement`

src/components/bottomSheet/types.d.ts
```diff
@@ -140,7 +141,23 @@ export type BottomSheetProps = {
...
+  /**
+   * Callback on snap position change that announces the returned string on screen readers.
+   * @type string | ((positionInScreen: string) => string)
+   */
+  accessibilityPositionChangeAnnouncement?:
```
gorhom, line 156: why not just to having a prop for text with placeholder like: 
```ts
// an announcement text with sheet change position
// example: `Bottom sheet snapped to POSITION% of the screen`
accessibilityPositionAnnouncement?: string
```

src/components/bottomSheet/types.d.ts
```diff
@@ -4,6 +4,7 @@ import type { State } from 'react-native-gesture-handler';
 import type { BottomSheetHandleProps } from '../bottomSheetHandle';
 import type { BottomSheetBackgroundProps } from '../bottomSheetBackground';
 import type { BottomSheetBackdropProps } from '../bottomSheetBackdrop';
+import { NullableAccessibilityProps } from '../../types';
```
gorhom: update to 
```
import type { NullableAccessibilityProps } from '../../types';
```

src/components/bottomSheetBackdrop/types.d.ts
```diff
@@ -1,5 +1,6 @@
 import type { ViewProps } from 'react-native';
 import type Animated from 'react-native-reanimated';
+import { NullableAccessibilityProps } from '../../types';
```
gorhom: update to
```
import type { NullableAccessibilityProps } from '../../types';
```

src/components/bottomSheetHandle/BottomSheetHandle.tsx
```diff
@@ -2,10 +2,32 @@ import React, { memo } from 'react';
 import { View } from 'react-native';
 import isEqual from 'lodash.isequal';
 import { styles } from './styles';
+import { BottomSheetDefaultHandleProps } from './types';
```
gorhom: add `type` ^

gorhom (changes requested):

src/types.d.ts
nmuranovas: There are some use cases where the user would like to set other accessibility props as well (e.g. accessibilityState, accessibilityValue, onAccessibilityTap, accessibilityActions)

src/components/bottomSheet/types.d.ts
nmuranovas: I think that having a `@type string | ((positionInScreen: string) => string)` is a bit easier to use but let me know if you would really prefer the placeholder approach :)

nmuranovas:
I have implemented most of the suggested changes and responded to the remaining ones 🙂

Please take a look when You have the time @gorhom 👍

gorhom:
thanks @nmuranovas for updating this PR, i will look into it later today 👍

src/types.d.ts
gorhom: okay , then why you redefined `accessible`, `accessibilityLabel`... , they are already in `AccessibilityProps` ?

src/components/bottomSheet/types.d.ts
gorhom: it's alright

src/types.d.ts
nmuranovas: That is to make them overwritable. To avoid breaking changes it will work the same for old users but those who want the screen reader to not read out these properties at all can pass a null

gorhom:
@nmuranovas  everything looks fine, i just wanted to understand why do we need `NullableAccessibilityProps` interface ? the `AccessibilityProps` interface variables are already optional ?

src/types.d.ts
gorhom: i think `undefined` should work fine ? right

nmuranovas: I made it to apply default values if nothing is defined (this avoids breaking changes)

Right now if property is undefined = apply default value
If property is null = apply undefined value

nmuranovas: If i pass undefined as a value it eill still apply default english values

gorhom: ah, okay

gorhom (approved):

PR #581, opened 2021-08-17
Update BottomSheetBackdrop.tsx

TheNerdGuyLulu:
Fix for TypeError: null is not an object (evaluating 'containerRef.current.setNativeProps')

## Motivation

Captura de ecrã 2021-08-17, às 02 17 00


gorhom:
this should be fixed with `v2.4.1`

Issue #578, opened 2021-08-15
[v4.0.0-alpha.27] When press on background bottom sheet becomes unresponsive


likern:
# Bug

https://user-images.githubusercontent.com/842306/129493938-d48a0c8c-b770-4b36-925f-dc5bd52ee866.mp4

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet         | v4.0.0-alpha.27   |
| react-native                    | 0.64.0   |
| react-native-reanimated         |  2.3.0-alpha.2  |
| react-native-gesture-handler    | 1.10.3   |

## Steps To Reproduce

1. Open bottom sheet (with `[-1, "CONTENT_HEIGHT"]`) snap points
2. Close it by pressing on backdrop component
3. Bottom sheet becomes unresponsive 

Describe what you expected to happen:

1. Bottom sheet should be responsive and I able to press buttons again

likern:
@gorhom I've made some investigation and this is what I've found.

`onLayout` of `BottomSheetContainer` shows this **height** and it's the value of `animatedClosedPosition.value`
```bash
[BottomSheetContainer::handleContainerLayout] height:768.727294921875
```
while dynamic animated snap points are somehow shifted **by 1 pixel**:
```bash
[BottomSheet::useAnimatedReaction::OnMount] nextPosition:769.727294921875 animatedSnapPoints:769.727294921875,390.1818542480469 isLayoutCalculated:true
```

When pressing on backdrop there is position **768.727294921875**
```
[BottomSheet::handleClose] 
[BottomSheet::animateToPosition] velocity:0 position:768.727294921875 currentPosition:390.1818542480469
```
this leads to `[BottomSheet::useAnimatedReaction::OnChange] ...` not being called and this line is not executed
https://github.com/gorhom/react-native-bottom-sheet/blob/b65ea0df0de2bd0e6fca35a96a7ed2b4c2d81cb5/src/components/bottomSheet/BottomSheet.tsx#L1408 

I think this is the reason.

It's not being called because the last `_animatedIndex.value` is `0.02` and this condition is `true`

https://github.com/gorhom/react-native-bottom-sheet/blob/b65ea0df0de2bd0e6fca35a96a7ed2b4c2d81cb5/src/components/bottomSheet/BottomSheet.tsx#L1375-L1377

likern:
```bash
[...]

PR #584, opened 2021-08-17
Allow usage of index when using dynamic snap points

### LINKED ISSUE #586: [v4.0.0-alpha.27] Exception when correct index is provided
# Bug

When correct `index={1}` is provided, when snap points are `[0, 'CONTENT_HEIGHT']` exception is thrown.

## Environment info

| @gorhom/bottom-sheet            | v4.0.0-alpha.27   |
| react-native                    | 0.64.0   |
| react-native-reanimated         | 2.3.0-alpha.2   |
| react-native-gesture-handler    | 1.10.3   |

## Steps To Reproduce

1. Open provided snack code with reproduction
2. Exception should be thrown

1. No exception should be thrown

## Reproducible sample code

https://snack.expo.dev/@victormalov/dynamicsnappoints-error

https://user-images.githubusercontent.com/842306/129749498-ee9234b7-09f7-4e6e-83e4-1cff8bf72fbd.mp4

Ferossgp:
Because the number of snap points is different on the initial frame with the number when the sheet size is known, the bottom sheet could not be used with an index value. To avoid this issue we fill the snap points array with INITIAL_SNAP_POINT, thus snapPoints array will have the same length before the contentHeight is available.

### DIFF
--- a/src/hooks/useBottomSheetDynamicSnapPoints.ts
+++ b/src/hooks/useBottomSheetDynamicSnapPoints.ts
@@ -29,7 +29,7 @@ export const useBottomSheetDynamicSnapPoints = (
       animatedHandleHeight.value === INITIAL_HANDLE_HEIGHT ||
       animatedContentHeight.value === 0
     ) {
-      return initialSnapPoints.map(_ => INITIAL_SNAP_POINT);
+      return initialSnapPoints.map(() => INITIAL_SNAP_POINT);
     }
     const contentWithHandleHeight =
       animatedContentHeight.value + animatedHandleHeight.value;
--- a/src/hooks/useBottomSheetDynamicSnapPoints.ts
+++ b/src/hooks/useBottomSheetDynamicSnapPoints.ts
@@ -29,7 +29,7 @@ export const useBottomSheetDynamicSnapPoints = (
       animatedHandleHeight.value === INITIAL_HANDLE_HEIGHT ||
       animatedContentHeight.value === 0
     ) {
-      return [INITIAL_SNAP_POINT];
+      return initialSnapPoints.map(_ => INITIAL_SNAP_POINT);
     }
     const contentWithHandleHeight =
       animatedContentHeight.value + animatedHandleHeight.value;


likern:
@Ferossgp Could you explain a little bit more, with use cases?

I think this PR is not correct. Initial `[INITIAL_SNAP_POINT]` is not a real array of snap points. It's a special sign for the rest of the code to handle it differently as _"Do not show bottom sheet because we don't know layouts"_.

That `[INITIAL_SNAP_POINT]` is used throughout the rest of the code. It's not enough to change it just there. You should change  logic inside `BottomSheet`.

This is my little understanding.

Ferossgp:
@likern 
> is not a real array of snap points

As far as I see it is used as a real array of snap points. It is passed to the `snapPoints` [here](https://github.com/gorhom/react-native-bottom-sheet/blob/master/example/src/screens/advanced/DynamicSnapPointExample.tsx#L62). If we add `index={1}`, meaning that we want the bottom sheet to open at the dynamic position when it is available, the [invariant will fail](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L723-L728). It will fail because the length is called on the snapPoints property which is `[INITIAL_SNAP_POINT]`.

> That [INITIAL_SNAP_POINT] is used throughout the rest of the code

In the rest of the code, [it uses only the value in array at index 0](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L246). Thus seems like there is no need for changes in other places.

The other way to fix it could be making the invariant aware of `isSnapPointsNormalized`, but this may complicate the logic of the bottom sheet component.

I'm happy to update the places where it should be adjusted if there are such.


Thanks!

likern:
Could you provide snack example which shows what this PR solves?

Might be I don't understand something.

Ferossgp:
Sure, https://snack.expo.dev/@ferossgp/dynamicsnappoints-error

likern:
> As far as I see it is used as a real array of snap points. It is passed to the snapPoints here.

The logic for treating this array **as special** in hidden inside `BottomSheet` component inself, internally.

> In the rest of the code, [it uses only the value in array at index 0](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L246). Thus seems like there is no need for changes in other places.

Yes, because it's a **special array**. It's not used to really draw bottom sheet on `-999` position, it's not a real snap point. It's used to not show bottom sheet until layout is calculated (`isLayoutCalculated` set to `true`). Then the logic is propogated through `isLayoutCalculated`.

likern:
I mean this PR might fix this error, but it doesn't correct in terms how the overall `BottomSheet` works internally. It becomes more and more fragile with hacks.

likern:
> Sure, https://snack.expo.dev/@ferossgp/dynamicsnappoints-error

Seems like a bug. Could you create this bug?

I think this `useBottomSheetDynamicSnapPoints` and it's logic was not tested carefully.

likern:
I created issue https://github.com/gorhom/react-native-bottom-sheet/issues/586

src/hooks/useBottomSheetDynamicSnapPoints.ts
```diff
@@ -29,7 +29,7 @@ export const useBottomSheetDynamicSnapPoints = (
       animatedHandleHeight.value === INITIAL_HANDLE_HEIGHT ||
       animatedContentHeight.value === 0
     ) {
-      return [INITIAL_SNAP_POINT];
+      return initialSnapPoints.map(_ => INITIAL_SNAP_POINT);
```
gorhom: i would suggested to re-write it this way

```tsx
      return initialSnapPoints.map(() => INITIAL_SNAP_POINT);
```

gorhom (changes requested):
thanks @Ferossgp , @likern  for submitting and reviewing this pr, the issue is valid and the fix should solve it.

src/hooks/useBottomSheetDynamicSnapPoints.ts
likern: > i would suggested to re-write it this way
> 
> ```tsx
>       return initialSnapPoints.map(() => INITIAL_SNAP_POINT);
> ```

I'm working on another implementation using `Symbol()` as `INITIAL_DYNAMIC_SNAP_POINTS` to pass to `snapPoints` parameter. It should solve #586 #578.

likern: @gorhom Or should I abandon this work in preference of this PR?

gorhom: @likern i have resolved #578 with https://github.com/gorhom/react-native-bottom-sheet/commit/95ea72a459f96d40ad583c5579cc72f0e128e5dd, and #586 should be resolved with this pr.

I tried to implement it using `Symbol` as initial values, but it will require a lot of refactoring to internal functionality.

gorhom: this pr is needed when using `useBottomSheetDynamicSnapPoints`, i was returning only one value `INITIAL_DYNAMIC_SNAP_POINTS` but internally i was checking for the snap point with provided index. which will cause the crash.

likern: @gorhom Yes, of course I'm talking about correct PR which will take into account this.

likern: Something like this.
Снимок экрана от 2021-08-18 00-18-55

likern: @gorhom If 🆗 I will continue implementation. In that case there will be no confusion and accident errors between 
`[-999]` special array and normal array for snap points. No confustion in parameter validations (don't allow negative values - now I get exception because first is rendered `-999` value). It's kind of a hacks which are layering on top of each other.

I think best way is to use symbols to distinguish from normal arrays.

gorhom: @likern what will be the different from this approach ? you'll still need to handle all places `_providedIndex` is referenced to handle one item array `_providedSnapPoints` & `animatedSnapPoints`. for me this approach is easier and correct my mistake on the initial commit for this hook.

i would like to hear your opinion :)

gorhom: @likern with the latest commit i pushed, the `snapPoints` will need to be `> 0 value` or an initial snap point `-999`.

>No confustion in parameter validations (don't allow negative values - now I get exception because first is rendered -999 value). It's kind of a hacks which are layering on top of each other.

I have tested this locally on dynamic snap point example. it should not throw an exception.

likern: For getting unique symbol (I mean ES6 Symbol type) user need

- Call `useBottomSheetDynamicSnapPoints` which will return that unique symbol 
It's not possible to accidentally pass `[-999]` special array to `snapPoints` parameter. I think it's more evident what's happening and why.

- When logging and debugging it's difficult to understand what is `[-999]` and what's going on. I initially thought it's a bug, untill I dig into code.
- While now that approach might work, I'm sure there will be more bugs. Because `[-999]` and user's array are just arrays with _valid values_ (from JS perspective) and it's difficult to distinguish them. While array and symbol are completely different (and unique, we have to use only one `INITIAL_DYNAMIC_SNAP_POINTS`, while can spread out `[-999]` directly)
- TypeScript supports `unique symbol` type allow differentiate easily (as in runtime in JS). Now we need some hacks - for example check that first value from array is `-999`. It feels like a hack and that information will be lost later while refactoring and changing code base.

gorhom (approved):

src/hooks/useBottomSheetDynamicSnapPoints.ts
likern: It doesn't solve problem that we need to check `animatedSnapPoints` in every place where we use it. I agree with that.

 - But you can't accidentally call `slice()` on `symbol | number[]`, while now you can. And there is some weird calculation (and probably wrong)

Also this code is correct and totally valid, except this weird `[-999]` special handle.
```ts
        invariant(
          index >= -1 && index <= snapPoints.length - 1,
          `'index' was provided but out of the provided snap points range! expected value to be between -1, ${
            snapPoints.length - 1
          }`
        );
```

likern: @gorhom I'm OK with any solution that's working ;) Thanks for fast fixing. Will check tomorrow.

PR #635, opened 2021-09-10
fix: remove 'removeListener' as it is now deprecated

brianathere:

## Motivation

removeListener is deprecated, and using it at runtime results in a handled exception. Converting to using the remove() method on the EvenSubscription returned from the addListener call eliminates this handled exception.


gorhom:
@brianathere thanks for submitting this pr 👏

PR #656, opened 2021-09-22
React on keyboard height change

### LINKED ISSUE #655: [v4] Bottom sheet does not react on keybaord height change
# Bug

The bottom sheet does not react to keyboard height changes. Because of that when the keyboard height is change by switching layout the keyboard covers the bottom sheet and makes the content on it unaccessible.

## Environment info

| @gorhom/bottom-sheet            | 4.0.3   |
| react-native                    | 0.65.0  |
| react-native-reanimated         | 2.2.1   |
| react-native-gesture-handler    | 1.10.1  |

## Steps To Reproduce

Open keyboard on iOS, switch to the emoji picker which has a bigger height than the English US layout. See that the difference in height covers the content.

When keyboard height changes the bottom sheet component should adapt and snap to the new change.

## Reproducible sample code

https://snack.expo.dev/@ferossgp/keyboard-handling-issue

https://user-images.githubusercontent.com/5794620/134296642-649bea93-fded-4a50-941e-cab8f125da60.MP4

Ferossgp:
Resolves #655

Keyboard state property changes only when the keyboard is opened or closed. But it does not react to changes that happen when the keyboard is already open. An example of such change is switch keyboard layout, or opening the emoji keyboard.

## Motivation

By reacting to height change we solve the issue when the keyboard changes the height from a smaller to a bigger one. In such cases, the keyboard covers the content disallowing the user to interact with that part of the bottom sheet.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -1333,8 +1333,15 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
      * @alias OnKeyboardStateChange
      */
     useAnimatedReaction(
-      () => animatedKeyboardState.value,
-      (_keyboardState, _previousKeyboardState) => {
+      () => ({
+        _keyboardState: animatedKeyboardState.value,
+        _keyboardHeight: animatedKeyboardHeight.value,
+      }),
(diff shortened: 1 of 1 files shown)


gorhom:
thanks @Ferossgp for submitting this pr <3

PR #669, opened 2021-09-29
fix: prototype pollution in immer by upgrading to @gorhom/portal v1.0.9

calintamas:
## Motivation

Opening a very small PR that addresses a security vulnerability found in v2, explained in more detail below:

- `react-native-bottom-sheet v2.4.0` uses `@gorhom/portal v1.0.4`
- `@gorhom/portal v1.0.4` depends on `immer v8.0.1`
- all immer versions under `v9.0.6` suffer from this [Prototype Pollution vulnerability](https://github.com/advisories/GHSA-33f9-j839-rf8h)

This PR bumps `@gorhom/portal` to `v1.0.9` which no longer uses immer, hence "resolving" the vulnerability.

@gorhom Thank you for all the work on the library!

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "bootstrap": "yarn install && yarn example"
   },
   "dependencies": {
-    "@gorhom/portal": "^1.0.4",
+    "@gorhom/portal": "^1.0.9",
     "invariant": "^2.2.4",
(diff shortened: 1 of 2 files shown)


gorhom (approved):
thanks @calintamas !

PR #676, opened 2021-10-01
feat: dismiss() without a key will remove the current modal if any

Shywim:
## Motivation

This enable calling `dismiss` from `useBottomSheetModal` without passing any key, so it can remove the currently presented `BottomSheetModal`, if any. It also now returns a boolean indicating whether a `BottomSheetModal` has been dismissed.

This enable users of this to call `dismiss` without having to manage themselves which modal is currently presented and can lead to a more straightforward usage of back handler on Android in most use-cases, for example :

```ts
const { dismiss } = useBottomSheetModal()
useBackHandler(dismiss)
```


gorhom (approved):
thanks @Shywim for submitting this PR

PR #690, opened 2021-10-09
fix: hide the sheet container on closed

### LINKED ISSUE #665: [v4] [Need Help] BottomSheet shows up for a brief second when dismissing keyboard on chat screen
# Bug

I'm looking for a suggestion with an issue I have. I'm using the latest version (4.1.0) of BottomSheet in my project and we are using it in our Chat screen.

In our Chat screen, when we click on the Input box, the Keyboard opens and the entire Chat screen lifts up.
At this point, the BottomSheet (which has an index of -1), is right behind the keyboard.

When we dismiss the Keyboard, the BottomSheet appears for a brief second because it was right behind it.

One solution I could think of was, providing a negative snapPoint, something like -500. This is exactly what we did with our previous BottomSheet library and it allowed us to provide it any negative snapPoint we want. So we didn't face this issue with that library.

When I tried to do the same with this library, it throws an error saying that negative snapPoints aren't allowed. I'm sure this isn't a bug. But I'd greatly appreciate any help or suggestion on this.

Please see the recording here :
https://user-images.githubusercontent.com/90186621/135208450-01f89b8c-4d28-4aad-b2e4-92019fbfde98.mp4

NOTE : It works well on iOS. The issue only exists on Android.

## Environment info

| @gorhom/bottom-sheet            | 4.1.0   |
| react-native                    | 0.63.3   |
| react-native-reanimated         | 2.2.0   |
| react-native-gesture-handler    | 1.10.3   |

## Steps To Reproduce

1. Add the BottomSheet in your Chat screen
2. Focus on the Input Box so that the Keyboard opens up and pushes the entire View up.
3. Dismiss the keyboard to see the BottomSheet appear for a fraction of a second.

Need advice on a workaround for this.

## Reproducible sample code

None. Please see the video recording.

gorhom:
Closes #665

## Motivation

When the user dismiss a keyboard view on Android, the container height will resize after the keyboard is been hidden, which will make the closed bottom sheet visible for few milliseconds. 

So i have added a logic to hide the bottom sheet when it is closed.

## Installation

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/665-hide-sheet-when-closed
```

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -1147,13 +1147,17 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
     //#endregion
 
     //#region styles
-    const containerAnimatedStyle = useAnimatedStyle(() => ({
-      transform: [
-        {
-          translateY: animatedPosition.value,
-        },
-      ],
-    }));
+    const containerAnimatedStyle = useAnimatedStyle(
+      () => ({
+        opacity: animatedIndex.value === -1 ? 0 : 1,
+        transform: [
+          {
+            translateY: animatedPosition.value,
+          },
+        ],
+      }),
+      [animatedPosition, animatedIndex]
+    );
     const containerStyle = useMemo(
       () => [_providedStyle, styles.container, containerAnimatedStyle],
       [_providedStyle, containerAnimatedStyle]


gorhom:
@aditya-keri-betterhalf, @siddharth-kt , @aimensasi would you mind testing this PR, thanks !

siddharthkr:
Great, I tested it, it worked for me 👍.

aimensasi:
@gorhom I confirm this is fixed the issue on my end. Thanks a lot

julienripet:
Hi, sorry to be a buzzkill, but it doesn't seem to fix it for me.  
I installed it using the command you gave above, `yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/665-hide-sheet-when-closed`, but it didn't do anything.  
I deleted my node_module folder, I restarted my metro server with `--reset-cache`, I even updated `react-native-reanimated` and `react-native-gesture-handler` to their latest versions, but nothing seems to change.

@siddharth-kt or @aimensasi, do you think you could share your package.json? Or if you have an idea of what I might be doing wrong?  

I can open another Issue with my code if needed.

aditya-keri-betterhalf:
I confirm as well that this issue is fixed on my end.

aimensasi:
@julienripet  I did not do anything special, just installed the fix and reloaded the app.

aimensasi:
@julienripet maybe your issue is different, you could create a snack to reproduce.

badalsaibo:
Was having this problem too. Ran `npm install ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/665-hide-sheet-when-closed` and it seems to be fixed.

gorhom:
thanks guys for testing!

rchavik:
I do not understand why, but this somehow broke the web version of my App.

I added `console.log({animatedIndex})` just before useAnimatedStyle() and debug console only shows 1 entry with animatedIndex.value = -1. So the opacity is 0.

I also tried to fish out the actual element via the Inspector and found the following:

image

Changing the opacity to 1 does show the bottom sheet again.

DanielMarkiel:
Hi @gorhom,
I believe this fix breaks the BottomSheetModal (it's probably the same bug as @rchavik described above).
It opens only once and it doesn't want to reopen after dismissing (everything is fine at v4.1.1).
My snap points (if this is relevant):
```
const initialSnapPoints = useMemo(() => ['CONTENT_HEIGHT'], [])
```

gorhom:
@rchavik could you open a new issue with a reproducible snack ? 

@DanielMarkiel could you check this issue https://github.com/gorhom/react-native-bottom-sheet/issues/697, it might be the one you are facing now

Kjagd:
@gorhom I'm also facing issues regarding this. Using `BottomSheetModal` sometimes when it's displayed for the second time, the opacity is not not reset, so it's completely invisible. It's a bit tricky to reproduce, since it apparently depends on the how much content I have in the sheet. But if I have something a bit heavy like a calendar in it, it's very consistently happening. Perhaps https://github.com/gorhom/react-native-bottom-sheet/issues/697 needs to be reopened.

PR #699, opened 2021-10-15
Use optional chaining to make sure that the ref exists

jcgertig:
## Motivation

Keep getting errors when I open a bottom sheet in a bottom sheet after its closed that there is no ref


gorhom:
thanks @jcgertig !

PR #724, opened 2021-10-27
feat: Add prop minDist for BottomSheet

Navipro70:
## Motivation

On screen where I'm use bottom sheet modal, gesture for back navigation is blocked in over 80% of attempts, because modal intercepts event. I sure that minDist will fix it:)


Navipro70:
Lol nice -_-

Issue #719, opened 2021-10-25
[v4] BottomSheetModal: Backdrop appears but Modal Content is not visible.


ryanhyslop:
# Bug
Very occasionally when calling present on a BottomSheetModal the backdrop will appear but the modal content will not be visible.

Oddly the content of the modal is still interact-able (i.e button presses get recorded) so the content is rendered, its just not visible (opacity: 0?).

This is very erratic, in that sometimes it happens first time, others it takes 5 or 10 trys dismissing and opening for it to occur. I have got it occurring in the snack below and have attached a screenshot to show.

I'm not ruling out us doing something wrong here and that snack is obviously a very reduced case but represents the broad use case. Primarily:
- `present()` called on button press
- rendered inside react navigation stack 
- BottomSheetFlatlist is inside the modal.

Just looking for pointers if we're doing something wrong and happy to help investigate and provide whatever information may be useful.

## Environment info

{
  "dependencies": {
    "@gorhom/bottom-sheet": "^4.1.3",
    "react-native-screens": "~3.4.0",
    "@react-navigation/stack": "^6.0.9",
    "react-native-reanimated": "~2.2.0",
    "@react-navigation/native": "^6.0.4",
    "react-native-gesture-handler": "~1.10.2",
    "react-native-safe-area-context": "3.2.0"
  }
}

## Steps To Reproduce

1. Tap 'Open model' 
2. Dismiss modal
3. Repeat open and close until eventually modal content doesn't appear.

Describe what you expected to happen:

1. Modal content reliably appears.
2.

## Reproducible sample code

Reminder that this doesn't happen straight away, can take 5 / 10 / 20 tries. Speed does not seem to matter though.

In addition we've only seemed to get this issue on iOS, not saying that it couldn't happen on Android only we haven't seen it yet.

https://snack.expo.dev/@ryanhyslop/bottom-sheet-v4-reproducible-issue-template

![IMG_2D42F22F3585-1](https://user-images.githubusercontent.com/323687/138688263-1676d7eb-fc3b-4f42-bba6-47da9aa0ff61.jpeg)

Thanks!

danilofatcat:
I am also experiencing this problem.
It seems like the content is actually there, just not visible. Even the pan down gesture is working, if you can guess where the handle component is.

sirajtahrawashmen:
I am also experiencing this problem. +1

ryanhyslop:
[...]

PR #730, opened 2021-10-29
Opacity animation through withTiming with 0 duration

ryanhyslop:
This explores solving issue: #719

Presently when toggling a modal the containerAnimatedStyle is setting opacity based on the animatedIndex value. Occasionally, and seemingly randomly, this fails and leaves the opacity set to 0, even if the modal is open. 

I don't have enough of an understanding of all the use cases to say whether this change breaks anything else but removing the opacity toggle entirely results in no observable problems and issue #719 no longer occurs.

If we did want to animate the opacity something like:
`opacity: interpolate(animatedIndex.value, [-1, 0], [0, 1])`

Would likely be more appropriate but in observing that it doesn't look desirable, but least works reliably.

This could be a short term patch for people having the same issue as us until someone with more experience on the project can advise.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -19,6 +19,7 @@ import Animated, {
   runOnUI,
   cancelAnimation,
   useWorkletCallback,
+  withTiming
 } from 'react-native-reanimated';
 import { State } from 'react-native-gesture-handler';
 import {
@@ -1149,6 +1150,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
     //#region styles
     const containerAnimatedStyle = useAnimatedStyle(
       () => ({
+        opacity: withTiming(animatedIndex.value === -1 ? 0 : 1, {duration: 0}),
         transform: [
           {
             translateY: animatedPosition.value,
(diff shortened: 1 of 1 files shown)


ryanhyslop:
Looking into it a bit further I note this PR introduced the opacity:
https://github.com/gorhom/react-native-bottom-sheet/commit/9f04d557d202ab8570b1b409332bfdd129e5efa4

To resolve this issue:
https://github.com/gorhom/react-native-bottom-sheet/issues/665

I can't quite seem to recreate the issue myself but seems a bit nuanced, if we wanted to keep the opacity but not have it transition something like this seems to work ok:

`opacity: withTiming(animatedIndex.value === -1 ? 0 : 1, {duration: 0}),`

That said I'm not sure why what's there presently doesn't work as there are examples in Reanimated documentation that show useAnimatedStyles applying static values from a ternary check on the `.value`. Perhaps a bug with Reanimated? Afraid I'm not experienced enough with either libraries to make call on this but guess the above may get someone out of a bind in the mean time.

A01123748:
@gorhom, this works!!!

A01123748 (approved):

dburdan:
Thanks for the patch! This appears to be working for us as well. We noticed the issue after upgrading to RN 0.66.x (and Reanimated to latest stable).

I'll post if I notice any adverse effects.

PR #740, opened 2021-11-11
feat: add useBottomSheetInternal_unsafe & useBottomSheetModalInternal…

jembach:
…_unsafe to access context

## Motivation

By using this package I discovered the problem of integrating the bottom sheet in combination with our TextInput component. While trying to add your supported solution for handling keyboard to our custom TextInput we run into the problem, that our app every time crashes, when our TextInput isn't rendered into a bottom sheet.

Therefore I have added two additional hooks which aren't checking if the context is present or not. Therefore i have called them unsafe. But with this solution we can check self if the context is present or not and react to each case.

The reason why I haven't updated the existing hooks is on the one hand that mostly everything in this package is relying on these hooks and on the other hand this could be cause for some user a breaking change.


renchap:
I was looking at something similar for the exact same use case.

Rather than adding a new hook, maybe we can use an argument to the current hook? It currently does not have any argument, so adding a new one should not cause any breakage.

jembach:
Hi @renchap,

at first time I also tried to implement using one hook with an param. But I struggled with the types. Hopefully this new version is fitting better. But please evaluate the types. I have used `as` because I haven't any other idea how to solve it that typescript don't have any problems.

renchap:
It should be cleaner to use Function Overloads like described here: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

Something like:
```ts
function useBottomSheetInternal(): BottomSheetInternalContextType
function useBottomSheetInternal(unsafe: true): BottomSheetInternalContextType
export function useBottomSheetInternal(unsafe: false): BottomSheetInternalContextType | null {
  const context = useContext(BottomSheetInternalContext);

  if (unsafe !== true && context === null) {
    throw "'useBottomSheetInternal' cannot be used out of the BottomSheet!";
  }

  return context;
}
```

I did not test it but it should be the way to do it properly.

jembach:
I have changed it to function overloading. should now be ready :)

likern:
@jembach @renchap I think this PR solves problem only partially. I have very similar issue when I try to use `BottomSheetBackdrop` component outside of `<BottomSheetModal />`.

`BottomSheetBackdrop` internally uses `useBottomSheet` hook https://github.com/gorhom/react-native-bottom-sheet/blob/d951a1976f5fd2e7a38bedbabb452a103b9644ea/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx#L36
and I see this bug
Screenshot_2021-11-14-17-02-45-823_com breeffy feeve

jembach:
@likern I support your problem to be solved. But your problem is affecting much more code than i thought about and will definitely cause a breaking change accross the hole app.

One solution could be to create a new global Context that provides the information about if it could be accessed unsafe. The other solution would be that everything could return null and it must be handled every time.

At least i cannot really understand the use case when you render the backdrop without the BottomSheetModal.

I would suggest to create another PR to solve your problem.

likern:
@jembach My use-case is described in this issue https://github.com/gorhom/react-native-bottom-sheet/issues/747. As soon as we use modal, we might have shared backdrop component, not tied to behaviour of one `<BottomSheet />`.

I just wanted to mention that this problem exists. I couldn't solve it easily, just will reimplement `Backdrop` from scratch.
I think generic and systematic solution would require redesign of library.

jembach:
Hi @gorhom 

how does it look to merge this request? I'll require it to solve some issues in our app.

jembach:
Hi @gorhom, could you spent a look on this PR

renchap:
This would still be very useful!

renchap:
Still useful!

renchap:
Still useful :)

gorhom:
@jembach for submitting this pr!

gorhom (approved):

Issue #692, opened 2021-10-11
[v4]  BottomSheet goes below screen content with elevation on Android


itsramiel:
# Bug

### Issue
on **Android**
I have a screen that has a component wrapped in a View that has elevation set to 2 in order to create a shadow; however, when I add a BottomSheet component to the same screen, the component with elevation of 2 covers the modal. I am trying to set elevation 3 on the BottomSheet using the style prop, but it is not working

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.1.1   |
| react-native                    | 0.63.2   |
| react-native-reanimated         | 2.2.0   |
| react-native-gesture-handler    | 1.10.2   |

## Steps To Reproduce

1. Create a Simple Screen
2. Add a component and set its elevation to anything above 0
3. Add the Bottomsheet

Describe what you expected to happen:

1. I want the BottomSheet to go over the content, not under it

## Reproducible sample code

```
import React, { useMemo, useRef, useCallback } from 'react';
import { View, StyleSheet } from 'react-native';
import BottomSheet, { enableLogging } from '@gorhom/bottom-sheet';
import List from './components/List';
import SearchHandle from './components/SearchHandle';
import Button from './components/Button';

enableLogging();

const App = () => {
  // refs
  const bottomSheetRef = useRef(null);

  // variables
  const snapPoints = useMemo(() => ["80%"], []);

  // callbacks
  const handleExpandPress = useCallback(() => {
    bottomSheetRef.current?.expand();
  }, []);
  const handleCollapsePress = useCallback(() => {
    bottomSheetRef.current?.collapse();
  }, []);
  const handleClosePress = useCallback(() => {
    bottomSheetRef.current?.close();
  }, []);

  return (
    <View style={styles.container}>
    <View style={{elevation:1, backgroundColor:'blue'}}>
      <Button onPress={handleExpandPress} title="Expand" />
      <Button onPress={handleCollapsePress} title="Collapse" />
      <Button onPress={handleClosePress} title="Close" />
    </View>
      <BottomSheet
        ref={bottomSheetRef}
        snapPoints={snapPoints}
        keyboardBehavior="interactive"
        handleComponent={SearchHandle}
        animateOnMount={true}
        android_keyboardInputMode="adjustPan">
        <List type="FlatList" />
      </BottomSheet>
    </View>
  );
};

const styles = StyleSheet.create({
  container: {
    flex: 1,
[...]

Issue #398, opened 2021-04-11
adding an elevation prop to BottomSheet to elevate the BottomSheetContainer


HellNSab:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@gorhom/bottom-sheet@3.6.3` for the project I'm working on.

I had a problem with elevated components showing up above the bottom sheet even if i added an elevation to the style of bottom-sheet, i figured we could add an elevation prop that would modify the BottomSheetContainer style directly and add this prop to the styling. (I'm using v3.6.3 by the way)

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@gorhom/bottom-sheet/lib/typescript/components/bottomSheet/types.d.ts b/node_modules/@gorhom/bottom-sheet/lib/typescript/components/bottomSheet/types.d.ts
index e2a04c2..8bb7aee 100644
--- a/node_modules/@gorhom/bottom-sheet/lib/typescript/components/bottomSheet/types.d.ts
+++ b/node_modules/@gorhom/bottom-sheet/lib/typescript/components/bottomSheet/types.d.ts
@@ -112,6 +112,12 @@ export interface BottomSheetProps
    * - `restore`: restore sheet position.
    */
   keyboardBlurBehavior?: keyof typeof KEYBOARD_BLUR_BEHAVIOR;
+  /**
+   * Enable elevation on bottom sheet container.
+   * @type number
+   * @default 0
+   */
+   elevation?: number;
 
   /**
    * View style to be applied at the sheet container,
diff --git a/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx b/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
index 6fcefa2..0e8077e 100644
--- a/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
+++ b/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
@@ -109,6 +109,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       containerHeight: _providedContainerHeight,
       topInset = 0,
       overDragResistanceFactor = DEFAULT_OVER_DRAG_RESISTANCE_FACTOR,
+      elevation = 0,
 
       // animated callback shared values
[...]

Issue #387, opened 2021-04-04
Elevation issue


mrsilph:
## Ask your Question

Hi! How can I set the elevation of the bottom sheet?

As you can see in the photo, some components that are under the sheet but have and elevation value of 5 (for shadows in android) are been drawn on the sheet, that isn't desirable. Thanks!

![Capture](https://user-images.githubusercontent.com/47247073/113522282-7cbc8380-959f-11eb-9fa8-5316a0a45d08.PNG)

gorhom:
hi @darmane, i will add a new guide to the documentations soon 👍

gorhom:
guide been added https://gorhom.github.io/react-native-bottom-sheet/adding-shadow

mrsilph:
Thank you! I'll check it tomorrow

safeer-ahmed:
@gorhom still its not working in Android

virus2016:
Don't understand how to add a shadow using the guide, no code example...

gorhom:
@virus2016 https://github.com/gorhom/react-native-bottom-sheet/blob/master/example/src/screens/advanced/ShadowExample.tsx

mmmoussa:
Hi @gorhom, I believe you have misunderstood the issue that is being reported here, and is also reported in other issues such as #398 and #692. The problem is not adding a shadow to the bottom sheet as you've provided an example for, but rather, having the bottom sheet appear above elements that should be below it but make use of the `elevation` prop on Android. Android lays elements with higher elevation on top of elements with lower elevation, and since the bottom sheet has low elevation by default (that is not configurable), it breaks whenever you need to render components on the screen behind it that need an elevation greater than 0. This is what's being shown in @darmane's screenshot. The +/- counter component and the weekday circles are not part of the bottom sheet. They are supposed to be rendered on the screen *behind* the bottom sheet. This is easily reproducible by drawing a component on the screen with a non-zero elevation and then opening the bottom sheet to cover it on Android.

erickreutz:
@gorhom I follow your guide for adding a shadow to `BottomSheetModal1 and my logs are littered with `WARN  (ADVICE) View #1093 of type RCTView has a shadow set but cannot calculate shadow efficiently. Consider setting a solid background color to fix this, or apply the shadow to a more specific component.` is this expected?

PR #743, opened 2021-11-11
feat: support Android elevation for BottomSheet component

fiznool:
## Motivation

When rendering a component in the top portion of the screen (i.e. outside of the `BottomSheet` component), if this or any child components include an `elevation` style property, they are rendered on top of the bottom sheet. This is concisely described in #387 and #692.

A fix was proposed in #398 but was closed due to not adhering to the issue template. The bulk of the changes in this PR follow this proposal: an `elevation` prop has been added to the `BottomSheet` component, allowing it to be elevated above any other component that may otherwise render on top of the sheet.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -124,6 +124,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       containerOffset: _providedContainerOffset,
       topInset = 0,
       bottomInset = 0,
+      elevation = 0,
 
       // animated callback shared values
       animatedPosition: _providedAnimatedPosition,
@@ -1566,6 +1567,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
               topInset={topInset}
               bottomInset={bottomInset}
               detached={detached}
+              elevation={elevation}
             >
               <Animated.View style={containerStyle}>
                 <BottomSheetBackgroundContainer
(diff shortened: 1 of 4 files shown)


crutchcorn:
Ideally, we should have the ability to style this IMO, with more than elevation. For example, I want to use `zIndex` instead of `elevation`

nidelson:
Interesting!

pistonsky:
I tried to set a shadow on android for the modal using `elevation`, but the modal is blank. The content is rendered underneath it, and even setting `elevation` on the content's View doesn't help. Any ideas?

zholmes1:
This prop works and this commit should be merged. Without it there is no other way to force the bottom sheet to appear over other absolute positioned items on the screen. @gorhom please merge.

@fiznool Thank you so much for this solution.

Anyone else, if you are looking to incorporate this solution right now, simply use patch-package to make the small changes in this commit.

https://www.npmjs.com/package/patch-package

https://github.com/studiozeffa/react-native-bottom-sheet/commit/c78eeab211991615595026b091b755286028d7a0

horva:
Waiting for this PR to be merged, too! Hope it will land soon. Thanks @fiznool for the fix.

eliteWalrus4164:
I also would like to see it merged! Thanks @fiznool

chuchuva:
@gorhom Please review and merge 🙏

saumyabhatnagar2000:
@gorhom Please Merge

crutchcorn:
This is not stale.

janroures:
any news on this?

adrianorazzini:
@gorhom do we have any update around this issue?

PR #769, opened 2021-12-05
fix: resume animation on interruption

### LINKED ISSUE #632: [v4] handleUnmountSheet seems to fire twice in some scenarios
# Bug

We're encountering an issue where handleUnmountSheet seems to fire twice when quickly dismissing and presenting modals causing the sheetsQueue to be incorrect. This happens in real world scenarios when there is heavy bridge traffic occuring. The actual bug is that the first (lower) bottom sheet gets stuck open, you can see this in the video provided. There's other bugs I think this introduces such as preventing presenting future bottom sheets in some scenarios simply because the internal sheetsQueue is incorrect.

## Environment info

| @gorhom/bottom-sheet            | 4.0.3   |
| react-native-reanimated         | 2.2.0   |

## Steps To Reproduce

1. Reduce animationConfig speed via animationConfigs prop
2. .present() second BottomSheet
3. .dismiss() second BottomSheet
4. repeat 2 and 3 quickly

handleUnmountSheet should fire once per dismiss.

## Reproducible sample code

https://snack.expo.dev/Y_ZK1ozQ-

https://user-images.githubusercontent.com/2807897/132748996-9f9d51a6-1d3f-4b1c-9d68-c2f35c28b48e.mp4

gorhom:
Closes #632 #575

## Motivation

This PR will allow the sheet to resume the snap animation when user interrupted it.

This issue was noticeable on Android due to `Gesture Handler` platform handling to the event `onStart` will be fired immediately when user touch the sheet, while on iOS, it will only start when user start panning.

## Installation

```bash
yarn add ssh://git@github.com:gorhom/react-native-bottom-sheet#fix/632-resume-animation-on-interruption
```

### DIFF
--- a/src/components/bottomSheetDraggableView/BottomSheetDraggableView.tsx
+++ b/src/components/bottomSheetDraggableView/BottomSheetDraggableView.tsx
@@ -5,20 +5,19 @@ import {
   useBottomSheetGestureHandlers,
   useBottomSheetInternal,
 } from '../../hooks';
+import { GESTURE_SOURCE } from '../../constants';
 import type { BottomSheetDraggableViewProps } from './types';
 import { styles } from './styles';
 
 const BottomSheetDraggableViewComponent = ({
+  gestureType = GESTURE_SOURCE.CONTENT,
   nativeGestureRef,
   refreshControlGestureRef,
   style,
   children,
   ...rest
 }: BottomSheetDraggableViewProps) => {
-  // refs
-  const panGestureRef = useRef<PanGestureHandler>(null);
-
-  // hooks
+  //#region hooks
   const {
     enableContentPanningGesture,
     simultaneousHandlers: _providedSimultaneousHandlers,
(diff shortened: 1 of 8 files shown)


likern:
@gorhom I've tested and couldn't reproduce _freeze_ of bottom-sheet. I think it's been solved. Thank you very much!

If we could have `duration` it would be easier to reproduce bugs (for slowing animations), like here https://github.com/software-mansion/react-native-reanimated/issues/2691.

likern:
> This issue was noticeable on Android due to Gesture Handler platform handling to the event onStart will be fired immediately when user touch the sheet, while on iOS, it will only start when user start panning.
 
Is this bug in RNGH? Maybe to submit to there?

gorhom:
> @gorhom I've tested and couldn't reproduce _freeze_ of bottom-sheet. I think it's been solved. Thank you very much!
> 
> If we could have `duration` it would be easier to reproduce bugs (for slowing animations), like here [software-mansion/react-native-reanimated#2691](https://github.com/software-mansion/react-native-reanimated/issues/2691).

you could use:
```tsx
const animationConfigs = useBottomSheetTimingConfigs({
  duration: 5000
})
```

gorhom:
@likern thanks for testing 👍

SaeedZhiany:
@gorhom 

I'm using `v4.1.5` and I'm still facing that problem. is it any config need to be set?

SaeedZhiany:
@gorhom 

I'm using `v4.1.5` and I'm still facing that problem. is it any config needs to be set?

PR #825, opened 2022-01-28
Adjust closed position if container height changes on Android

nhannah:
## Motivation

On Android if you have a tall bottom sheet present but closed, then open the full keyboard, then select the options and one handed keyboard, the keyboard shrinks. This will then expose the handle of the hidden tall bottom sheet. It appears this is because the `animatedPosition` remains unchanged despite the keyboard shrinking and the `snapPoints` are unchanged `[0]`. Piggybacking on this `animatedReaction` we can check if the current `animatedPosition` is the same as the `_previousContainerHeight` then adjust the position to the new `containerHeight`. This solves the issue, I believe there will be no side effects as the additional checks keep this reassignment from occurring when other functions could change the `animatedPosition` value.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -1286,6 +1286,14 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
           !isLayoutCalculated.value ||
           !isAnimatedOnMount.value
         ) {
+          if (
+            Platform.OS === 'android' &&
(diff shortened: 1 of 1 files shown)


virdesai (approved):

nhannah:
I don't think this should have auto closed.

DeniferSantiago:
I think this will help me with my case, when I open the keyboard my BottomTabNavigation is hidden but my BottomSheet only handles the resize of the keyboard not the change of my BottomTab this causes part of the BottomSheet to be visible.

Can I help with something?

EnzoPortela:
That would be very helpful! @gorhom

PR #889, opened 2022-03-24
chore: adds the support to customize accessibility values

CrisTofani:

## Motivation

This PR aims to include the changes of #571 on master branch to be supported even on v4.

It adds the possibility to customize accessibility attributes from users integrating bottom-sheets on their projects.

In addition to base changes proposed by #571 this makes possible to include custom attributes on BottomSheetHandleContainer when declaring BottomSheet parent component.


CrisTofani:
I think this is still valid?

Issue #862, opened 2022-02-25
[v4] | [v2] iOS BottomSheet onClose event keep triggering after bottomsheet closed 


bayupuspanugraha:
# Bug

iOS BottomSheet onClose event keep triggering after bottomsheet closed 
   
I am using onClose event to hide backdrop, not sure why the backdrop is not automatically closed when bottomsheet is closed. (maybe team can have a solution for this, so no need manual trick to hide it at the future).
           
## Environment info 

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | ^4   |
| react-native                    | 0.66.0   |
| react-native-reanimated         | ^2.4.1   |
| react-native-gesture-handler    | ^2.2.0  |

## Steps To Reproduce
  
1. Open BottomSheet by using `bottomSheetRef.current?.expand()` 
2. Then Close the BottomSheet by using `bottomSheetRef.current.close()`,
3. When open another TextInputField on the background onClose bottom sheep keep triggering

Describe what you expected to happen:

1. If bottomsheet already closed, the onClosed event should not be triggered
 
  
## Snipped Code:

```
const [isShowModal, setIsShowModal] = useState(false)
const bottomSheetRef = useRef<BottomSheet>(null)

const renderBackdrop = useCallback(
        (props: any) => {
            const showBackDrop = isShowModal
            return (
                <BottomSheetBackdrop
                    {...props}
                    appearsOnIndex={1}
                    animatedIndex={{
                        value: 1,
                    }}
                    closeOnPress={true}
                    enableTouchThrough={true}
                    disappearsOnIndex={!showBackDrop ? 1 : 0}
                />
            )
        },
        [isShowModal],
    )

 const generateContent = () => (
        <View
            style={{
                padding: getSizePercentage(24),
                paddingTop: isIOS ? 0 : getSizePercentage(24),
                backgroundColor: WHITE,
                borderTopRightRadius: 10,
                borderTopLeftRadius: 10,
                height: getSizePercentage(maxHeight > -1 ? maxHeight : 250),
            }}
        >
            {isShowSearchBar && (
                <View style={{ marginBottom: getSizePercentage(5) }}>
                      {/* We wrap BottomSheetTextInput component here */}
                    <FormTextSearchInput
                        placeholder={`Search`}
                        value={searchKey}
                        onChangeText={(value) => setSearchKey(value)}
                        isIOS={isIOS}
                    />
                </View>
            )}

[...]

PR #901, opened 2022-04-04
Fix: onClose loop

chreck:
There is already an issue about the looping of triggering onClose https://github.com/gorhom/react-native-bottom-sheet/issues/862

## Motivation

Calling `expand()` and closing the bottom sheet by slide it down triggers `onClose` in a loop https://github.com/gorhom/react-native-bottom-sheet/blob/0010008906154f9a545f89d5826ea7af48336610/src/components/bottomSheet/BottomSheet.tsx#L1517


chreck:
I added aslo a fix for the situation when the user pulls down but releases finger after half of screen. It now sets animatedIndex to -1 manually.

gorhom:
thanks @chreck for submitting this pr, i tried to repro the issue on my side , but i couldn't :(

would you please provide a minimal repro sample code ?

chreck:
I commit an example with react native screens. It raises the onClose event, without my changes it did not. 
image

Cleversou1983:
I'm facing the same issue

PR #919, opened 2022-04-20
fix: Always update height to avoid races.

elan:
Updating a shared values is an asynchronous operation, so if two layouts happen in quick succession, we can miss an update.

## Motivation

On iOS devices, when backgrounding an app, the OS takes two screenshots (normal, and then rotated) so it can display thumbnail in the app switcher regardless of orientation. Apps see a rather confusing success of two screen rotations (and thus layouts).

What would happen is in some cases (it's a race), the container height would end up changing only once to the landscape height, so when you foregrounded the app again your displayed sheet would suddenly be cropped.

So here's the sheet showing properly (full screen player in Plexamp):

Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-04-20 at 10 59 40

After backgrounding and foregrounding, and missing a change to the shared value:

Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2022-04-20 at 11 00 11

### DIFF
--- a/src/components/bottomSheetContainer/BottomSheetContainer.tsx
+++ b/src/components/bottomSheetContainer/BottomSheetContainer.tsx
@@ -36,9 +36,7 @@ function BottomSheetContainerComponent({
         layout: { height },
       },
     }: LayoutChangeEvent) {
-      if (height !== containerHeight.value) {
-        containerHeight.value = height;
-      }
+      containerHeight.value = height;
 
       containerRef.current?.measure(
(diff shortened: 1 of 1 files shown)


elan:
Quote from the Reanimated docs:

> We can make updates to the state, but the updates are not immediate, and in order to read the data we need to wait till the next re-render.

gorhom:
@elan thanks for spotting this issue and submitting a PR !

PR #940, opened 2022-05-12
Allow BottomSheetContainer style overrides

natew:

## Motivation

Using this on an app and it appears behind other elements, can't see a way to get it above without changing this, unless I'm missing something.

### DIFF
--- a/src/components/bottomSheetContainer/BottomSheetContainer.tsx
+++ b/src/components/bottomSheetContainer/BottomSheetContainer.tsx
@@ -13,6 +13,7 @@ function BottomSheetContainerComponent({
   shouldCalculateHeight = true,
   detached,
   children,
+  containerStyle,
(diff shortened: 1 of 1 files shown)


gorhom:
hi @natew , could you explain more your use case ?

natew:
When I put BottomSheetProvider at the root of my app (on web), it creates a portal div somewhere in the DOM. But that div does not z-index above my main content. I can't control it currently without access to the style prop here, unless I'm missing something, so all my Bottom Sheets are appearing behind my app.

gorhom:
thanks @natew , i have pushed another commit to allow passing container style https://github.com/gorhom/react-native-bottom-sheet/commit/5e1ed9da98913d47b27912f49cf7e12b2393176e, it should be released on the next cut

PR #942, opened 2022-05-14
feat: added data to present modal api

gorhom:
## Motivation

This PR will introduce a new parameter to `present` from `useBottomSheetModal` hook, that will be passed the modal content.


gifuzzz:
is this feature not implemented yet? I can't find `present` in `useBottomSheetModal`.

`useBottomSheetModal` only has `dismiss` and `dismissAll`

PR #944, opened 2022-05-15
Allow "pressBehavior" to accept a function

pietroanello:

## Motivation

If the default BottomSheetBackdrop component, with pressBehavior 'close', is pressed before the BottomSheet is fully opened, "onAnimate" and "onChange" events are not fired. Anyone could have some context variables that changes when the state of the sheet change, and in this particular case I can only use the onClose event to detect the immediate closing of the sheet, losing a lot of performance.

In this way I can pass a function to the pressBehavior, making the setState and the close (triggered in my app with ref.current.close) events synchronized.


pietroanello:
Of course also having onAnimate triggered twice could be an option

PR #948, opened 2022-05-17
docs: Update getting-started.mdx

hirbod:
## Motivation

People still missing the fact that they have to wrap their app with a `GestureHandlerRootVIew` if upgraded to RNGH v2.
Let's just make it more obvious.


gorhom:
thanks @hirbod !

PR #963, opened 2022-05-28
Fix race condition when unmounting modal while opening it

elliottkember:
## Motivation

When a Bottom Sheet is animating in, its currentIndexRef is `-1`, and is not set to `0` until the animation finishes. This means that if the component is unmounted while it's still opening, the modal sticks around on screen and is uncloseable. This results in a zombie bottom sheet modal that can never be removed!  🧟

## To replicate:

```tsx
  const [showModal, setShowModal] = useState(false);
  useEffect(() => {
    setShowModal(true);
    setTimeout(() => {
      setShowModal(false);
    }, 200);
  }, []);

...

{showModal && <BottomSheetPullUp />}
```

## Changes:
- adds an `animatingRef` which tracks the status of the component's entry animation. 
- checks this `animatingRef` when minimizing the modal
- `animatingRef` is initialized as false `animateOnMount` is set to false!
- uses `onAnimate` to set `animatingRef = true` when animating

### DIFF
--- a/src/components/bottomSheetModal/BottomSheetModal.tsx
+++ b/src/components/bottomSheetModal/BottomSheetModal.tsx
@@ -73,6 +73,7 @@ const BottomSheetModalComponent = forwardRef<
   const bottomSheetRef = useRef<BottomSheet>(null);
   const currentIndexRef = useRef(!animateOnMount ? index : -1);
   const restoreIndexRef = useRef(-1);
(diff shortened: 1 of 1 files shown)


elliottkember:
I think this is still valid!

iqorlobanov:
Fixed here: https://github.com/gorhom/react-native-bottom-sheet/pull/2564

PR #962, opened 2022-05-27
Fix broken link for `Navigator Example`

fabriziocucci:
Fix broken link in `React Navigation Integration` page.

PR #965, opened 2022-05-29
Fix broken react-navigation example link

hirbod:
Fix broken example link.

PS: the file is called react-naigation, not a big deal but we should also change that to react-navigation :D


MANTENN:
https://github.com/gorhom/react-native-bottom-sheet/pull/966

hirbod:
Lol there is also another one: https://github.com/gorhom/react-native-bottom-sheet/pull/962

Guess we all stumbled on this without checking the PRs

Sydney-o9 (approved):

PR #979, opened 2022-06-09
fix(regression): updated keyboard handling reaction (by @yusufyildirim)

### LINKED ISSUE #946: [v4] | [v2] Keyboard does not automatically resize on the first try. Works after that though.
# Bug
The sheet does not auto resize the content height when keyboard is opened for the first time. It works after that but does not work if you clear the app and open the sheet with TextInput (BottomSheetTextInput) as a content in it for the first time. Video attached below:

## Environment info
| @gorhom/bottom-sheet            | "^4.3.0" |
| react-native                    | "^0.63.4" |
| react-native-reanimated         | "^2.4.1"|
| react-native-gesture-handler    | "~1.10.2" |

## Steps To Reproduce

1. Use the below provided BottomSheet as a standalone component.
2. Just use a Controlled TextInput as a children.
 3. Focus the input, the keyboard wont resize the content height when its visible.

1. The content height should move up the keyboard height when its visible everytime not only after the first time

## Reproducible sample code

`BottomSheet.js` component
```
import React, {useRef} from 'react';
import {BackHandler, useWindowDimensions} from 'react-native';
import {
  BottomSheetView,
  BottomSheetModal,
  BottomSheetBackdrop,
  useBottomSheetDynamicSnapPoints,
import {useSafeAreaInsets} from 'react-native-safe-area-context';

const BottomSheet = ({
  visible,
  setVisible = () => null,
  onDismiss = () => null,
  children = null,
  automaticallyAdjustHeight = false, // Prop 'key' with unique value as string needs to be passed to calculate the layout changes every time.
  backdropComponent = null,
  enableHandle = true,
  panToDismiss = true,
  responsive = true,
  snapPoints: _snapPoints = ['25%', '50%'],
  index = 1,
  modalStyle = {},
}) => {
  const bottomSheetRef = useRef(null);
  const {bottom: safeAreaBottom} = useSafeAreaInsets();

  const {window} = useWindowDimensions();

  const MAX_WIDTH = 600;
  const marginHorizontal =
    window.width > MAX_WIDTH ? (window.width - MAX_WIDTH) / 2 : 0;

  const snapPoints = automaticallyAdjustHeight
    ? ['CONTENT_HEIGHT']
    : _snapPoints;

  const {
    animatedHandleHeight,
    animatedSnapPoints,
    animatedContentHeight,
    handleContentLayout,
  } = useBottomSheetDynamicSnapPoints(snapPoints);

  React.useEffect(() => {
    if (visible) {
    } else {
      bottomSheetRef.current?.dismiss();
    }
  }, [visible]);

  const renderActualChildren = automaticallyAdjustHeight ? (
    <BottomSheetViewComponent
      handleContentLayout={handleContentLayout}
      key="actualChildren">
      {children}
    </BottomSheetViewComponent>
  ) : (
    children
  );

  const handleDismiss = React.useCallback(() => {
    setVisible(false);
    onDismiss?.();
  }, [onDismiss, setVisible]);

  const backPressed = React.useCallback(() => {
    setVisible(false);
    return true;
  }, [setVisible]);

  React.useEffect(() => {
    if (visible) {
      BackHandler.addEventListener('hardwareBackPress', backPressed);
    } else {
      BackHandler.removeEventListener('hardwareBackPress', backPressed);
    }
    return () => {
      BackHandler.removeEventListener('hardwareBackPress', backPressed);
    };
  }, [visible, backPressed]);

  return visible ? (
    <BottomSheetModal
      style={[
        {paddingBottom: safeAreaBottom},
        responsive ? {marginHorizontal} : {},
        modalStyle,
      ]}
      name="bottomsheet"
      ref={bottomSheetRef}
      enableHandlePanningGesture
      onDismiss={handleDismiss}
      enableContentPanningGesture={panToDismiss}
      keyboardBehavior="interactive"
      keyboardBlurBehavior="restore"
      backdropComponent={props => <Backdrop {...props} />}
      {...(automaticallyAdjustHeight ? {} : {index})}
      {...(backdropComponent !== null ? {backdropComponent} : {})}
      {...(!enableHandle ? {handleComponent: null} : {})}
      {...(automaticallyAdjustHeight
        ? {
            snapPoints: animatedSnapPoints,
            handleHeight: animatedHandleHeight,
            contentHeight: animatedContentHeight,
            key: `Modal${
              automaticallyAdjustHeight ? animatedContentHeight.value : ''
            }`,
          }
        : {snapPoints: _snapPoints})}>
      {renderActualChildren}
    </BottomSheetModal>
  ) : null;
};

function BottomSheetViewComponent({
  handleContentLayout = () => null,
  children = null,
}) {
  const {bottom: safeAreaBottom} = useSafeAreaInsets();

  return (
    <BottomSheetView
      onLayout={handleContentLayout}
      style={{
        paddingBottom: safeAreaBottom,
      }}>
      {children}
    </BottomSheetView>
  );
}

function Backdrop({...props}) {
  return (
    <BottomSheetBackdrop
      key="backdrop"
      {...props}
      pressBehavior={'close'}
      appearsOnIndex={0}
      disappearsOnIndex={-1}
    />
  );
}

export default BottomSheet;

```

Recording:
https://user-images.githubusercontent.com/51436631/168493392-6f4a5f24-213d-49e9-828f-08aa94b0da0a.mp4

yusufyildirim:
This PR fixes #946.

## Motivation

Looks like a regression is introduced with the PR https://github.com/gorhom/react-native-bottom-sheet/pull/931. With that PR, `getKeyboardHeightInContainer` renamed to `animatedKeyboardHeightInContainer` and started using `useDerivedValue` instead of `useWorkletCallback` which made it a `SharedValue` instead of a `function` to fix another bug. Yet introduced #946, a tricky bug. Here's the screen recording:

https://user-images.githubusercontent.com/22980987/172833171-beca69c8-2b15-4428-8aff-6e14465b9205.mov

Here's where I placed the `console.log`s to debug:

* Input focused: https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheetTextInput/BottomSheetTextInput.tsx#L17
* Keyboard Height Changed: https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/hooks/useKeyboard.ts#L55
* Handle Interactive Behavior: https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheet/BottomSheet.tsx#L589
* Keyboard state change detected: https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheet/BottomSheet.tsx#L1438
* `animatedKeyboardHeightInContainer` recalculated: https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheet/BottomSheet.tsx#L309

## Why it's happening?
There's [OnKeyboardStateChange reaction](https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheet/BottomSheet.tsx#L1376) that reacts to keyboard height and state changes to push the bottom sheet up when necessary. It calls [getNextPosition](https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheet/BottomSheet.tsx#L538) function (right [here](https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheet/BottomSheet.tsx#L1438)) to understand where the bottom sheet should be placed so it can animate it to that position. So far so good. `getNextPosition` function [depends on](https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheet/BottomSheet.tsx#L590) [`animatedKeyboardHeightInContainer` value](https://github.com/yusufyildirim/react-native-bottom-sheet/blob/fix/keyboard-handling/src/components/bottomSheet/BottomSheet.tsx#L297) --used to be `getKeyboardHeightInContainer`-- make the calculation. `animatedKeyboardHeightInContainer` derives itself from `animatedKeyboardHeight` which is pretty much the keyboard height value. This is where the bug is.

If you check the screen recording above carefully, you'll see on the second line that the keyboard height value is immediately set. But yet, on the third line, `animatedKeyboardHeightInContainer` value is `0` when it's accessed within the `getNextPosition` function. I said `animatedKeyboardHeightInContainer` is derived from the keyboard height, so "how's this possible?" you might ask. The answer is on the fifth line, `animatedKeyboardHeightInContainer` value is being recalculated *after* `OnKeyboardStateChange` reaction and `getNextPosition` function are called.

It used to be not a problem because `getNextPosition` was using `getKeyboardHeightInContainer` which was a synchronous function call. `getKeyboardHeightInContainer` was accessing the latest keyboard height value which was already set at the time it's called. Since it's replaced with `animatedKeyboardHeightInContainer`, a derived state, whenever we access it, we get the latest memoized result. If it's not recalculated before we call it, we don't actually get the latest state.

To fix this problem, I added `animatedKeyboardHeightInContainer` to the `OnKeyboardStateChange` reaction as an input hoping it'll be re-triggered even if `animatedKeyboardHeightInContainer` calculation occurs later like it happens in our case. But, while I was waiting for a re-trigger, it never happened. `animatedKeyboardHeightInContainer` started to be calculated before, all the time, which is good. I believe `reanimated` running some sort of dependency analysis algorithm that started to understand it should recalculate `animatedKeyboardHeightInContainer` before calling the `OnKeyboardStateChange`.

Quite a story for a one-line change but I think it is super important to explain. I hope it helps!

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -1379,6 +1379,8 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       () => ({
         _keyboardState: animatedKeyboardState.value,
         _keyboardHeight: animatedKeyboardHeight.value,
+        _animatedKeyboardHeightInContainer:
+          animatedKeyboardHeightInContainer.value,
       }),
       (result, _previousResult) => {
         const { _keyboardState, _keyboardHeight } = result;
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -291,36 +291,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       animationEasing: keyboardAnimationEasing,
       shouldHandleKeyboardEvents,
     } = useKeyboard();
-    /**
-     * Returns keyboard height that in the root container.
-     */
-    const animatedKeyboardHeightInContainer = useDerivedValue(() => {
-      /**
-       * if android software input mode is not `adjustPan`, than keyboard
-       * height will be 0 all the time.
-       */
-      if (
-        Platform.OS === 'android' &&
-        android_keyboardInputMode === KEYBOARD_INPUT_MODE.adjustResize
-      ) {
-        return 0;
-      }
-
-      return $modal
-        ? Math.abs(
-            animatedKeyboardHeight.value -
-              Math.abs(bottomInset - animatedContainerOffset.value.bottom)
-          )
-        : Math.abs(
-            animatedKeyboardHeight.value - animatedContainerOffset.value.bottom
-          );
-    }, [
-      $modal,
-      android_keyboardInputMode,
-      bottomInset,
-      animatedKeyboardHeight,
-      animatedContainerOffset,
-    ]);
+    const animatedKeyboardHeightInContainer = useSharedValue(0);
     //#endregion
 
     //#region state/dynamic variables
@@ -1379,13 +1350,21 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       () => ({
         _keyboardState: animatedKeyboardState.value,
         _keyboardHeight: animatedKeyboardHeight.value,
-        _animatedKeyboardHeightInContainer:
-          animatedKeyboardHeightInContainer.value,
       }),
       (result, _previousResult) => {
         const { _keyboardState, _keyboardHeight } = result;
         const _previousKeyboardState = _previousResult?._keyboardState;
         const _previousKeyboardHeight = _previousResult?._keyboardHeight;
+
+        /**
+         * Calculate the keyboard height in the container.
+         */
+        animatedKeyboardHeightInContainer.value = $modal
+          ? Math.abs(
+              _keyboardHeight -
+                Math.abs(bottomInset - animatedContainerOffset.value.bottom)
+            )
+          : Math.abs(_keyboardHeight - animatedContainerOffset.value.bottom);
 
         const hasActiveGesture =
           animatedContentGestureState.value === State.ACTIVE ||
@@ -1419,6 +1398,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
             keyboardBehavior === KEYBOARD_BEHAVIOR.interactive &&
             android_keyboardInputMode === KEYBOARD_INPUT_MODE.adjustResize)
         ) {
+          animatedKeyboardHeightInContainer.value = 0;
           return;
         }
 
@@ -1444,9 +1424,12 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
         );
       },
       [
+        $modal,
+        bottomInset,
         keyboardBehavior,
         keyboardBlurBehavior,
         android_keyboardInputMode,
+        animatedContainerOffset,
         getNextPosition,
       ]
     );


edkimmel (approved):

fa11erX (approved):

gorhom:
@yusufyildirim thanks for diving deep investigating this bug, and it is a regression ( my bad 🤦‍♂️ ). However your propose solution will trigger the reaction when the derived value changes, this may cause duplicate calls and other issues later on.

Thanks for your investigation, i have converted the `animatedKeyboardHeightInContainer` to a shared value that will be set on the keyboard height reaction, so its value will be updated instantly before executing any reaction.

@yusufyildirim could you please test my latest commit and confirm if it fixes the initial issue, thanks !

yusufyildirim:
> However your propose solution will trigger the reaction when the derived value changes, this may cause duplicate calls and other issues later on.

I did test that, wouldn't propose it as a solution if it was the case 😄 That's what I was trying to explain by saying;

> But, while I was waiting for a re-trigger, it never happened. animatedKeyboardHeightInContainer started to be calculated before, all the time, which is good. I believe reanimated running some sort of dependency analysis algorithm that started to understand it should recalculate animatedKeyboardHeightInContainer before calling the OnKeyboardStateChange.

It doesn't mean it won't happen, though. I'll check your commit when I find some time. Thanks!

yusufyildirim:
@gorhom Tested. It works!

gorhom:
thanks @yusufyildirim !

fa11erX:
Not Work corretly for me :(

3runoDesign:
Not Work corretly for me :(

PR #988, opened 2022-06-16
feat: add scrollable scroll position callback

cam-shaw:
## Motivation

At the moment, it's not possible with the default config to listen to the scroll position change as the onScroll prop is not passed down to the underlying ScrollView component.
This change allows us to listen to the scroll position change and do things such as animating the header component.

On scroll position update in the default scroll events handler, update scrollableContentOffsetY.
In `createBottomSheetScrollableComponent`, perform the callback and provide the latest scrollableContentOffsetY value through the use of `useAnimatedReaction`.

Currently I use this functionality in a patch-package, and it works for us.

**Example usage:**

```
<BottomSheetScrollView
    onScrollPositionUpdate={({position}) => {
        // Do something with the position value
    }}>
    ...
</BottomSheetScrollView>
```

### DIFF
--- a/src/components/bottomSheetScrollable/createBottomSheetScrollableComponent.tsx
+++ b/src/components/bottomSheetScrollable/createBottomSheetScrollableComponent.tsx
@@ -1,6 +1,11 @@
 import React, { forwardRef, useImperativeHandle, useMemo, useRef } from 'react';
 import { Platform } from 'react-native';
-import { useAnimatedProps, useAnimatedStyle } from 'react-native-reanimated';
+import {
+  runOnJS,
+  useAnimatedProps,
+  useAnimatedReaction,
+  useAnimatedStyle,
+} from 'react-native-reanimated';
 import { NativeViewGestureHandler } from 'react-native-gesture-handler';
 import BottomSheetDraggableView from '../bottomSheetDraggableView';
 import BottomSheetRefreshControl from '../bottomSheetRefreshControl';
@@ -35,6 +40,7 @@ export function createBottomSheetScrollableComponent<T, P>(
       style,
       refreshing,
       onRefresh,
+      onScrollPositionUpdate,
       progressViewOffset,
       refreshControl,
       ...rest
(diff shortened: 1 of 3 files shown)


cam-shaw:
@schamberg97 I think this would address your issue as mentioned 
https://github.com/gorhom/react-native-bottom-sheet/pull/952#issuecomment-1134568233

Aaron46:
How do you use this?

cam-shaw:
> How do you use this?
_________
Ah yes, good call out. Have added this to the main comment.
> **Example usage:**
> 
> ```
> <BottomSheetScrollView
>     onScrollPositionUpdate={({position}) => {
>         // Do something with the position value
>     }}>
>     ...
> </BottomSheetScrollView>
> ```

engee-nl:
Does this also work for BottomSheetFlatList?
Thank you!

cam-shaw:
> Does this also work for BottomSheetFlatList? Thank you!

It sure does! Works with all scrollable components

schamberg97:
> @schamberg97 I think this would address your issue as mentioned
> [#952 (comment)](https://github.com/gorhom/react-native-bottom-sheet/pull/952#issuecomment-1134568233)

After cursory glance at your code - this indeed appears to be exactly what we need, thank you 👍  Your solution is just more specific to our common problem, but according to @gorhom [here](https://github.com/gorhom/react-native-bottom-sheet/pull/952#issuecomment-1133647845) should perhaps be easier to maintain?

cam-shaw:
> > @schamberg97 I think this would address your issue as mentioned
> > [#952 (comment)](https://github.com/gorhom/react-native-bottom-sheet/pull/952#issuecomment-1134568233)
> 
> After cursory glance at your code - this indeed appears to be exactly what we need, thank you 👍 Your solution is just more specific to our common problem, but according to @gorhom [here](https://github.com/gorhom/react-native-bottom-sheet/pull/952#issuecomment-1133647845) should perhaps be easier to maintain?

Agree. I think if there is hesitation at this stage to expose the underlying hooks, then this could be a good alternative in the meantime. It would be a breaking change to remove it though, but probably manageable.

This looked like a decent alternative, but only if the hooks are exposed, so perhaps is more of a "future state" based on Gorhom's comment in that thread.
https://github.com/gorhom/react-native-bottom-sheet/pull/952#issuecomment-1161614075)

cam-shaw:
Pretty sure release 4.4.0 solves this issue (and in a much better way). Will close when I can give it a test in our project.

cam-shaw:
Release [4.4.0](https://github.com/gorhom/react-native-bottom-sheet/releases/tag/v4.4.0) addresses the issue this PR covers.
https://github.com/gorhom/react-native-bottom-sheet/pull/1019/files

Closing this PR.

PR #1019, opened 2022-07-04
feat: allow scrollable events

gorhom:
## Motivation

This PR will allow users to use `onScroll`, `onScrollBeginDrag` & `onScrollEndDrag` events on the scrollable components.

^ events should run on JS thread, no support for worklets, yet.


MichaelDanielTom:
@gorhom Is there any reason not to also include the same type of logic for `onMomentumScrollBegin` and `onMomentumScrollEnd`? If not, happy to put up a quick pr for that.

Separately, what are your thoughts on including parallel config for allowing worklets like `onScrollWorklet`, `onScrollBeginDrag`, etc.? It's not perfect but possibly better than trying to detect if a passed `onScroll` function is a worklet or anything like that. Right now I've been modifying the handlers in `useScrollHandler` (via patch-package) to allow for worklets, but that's pretty messy too.

MichaelDanielTom:
Kinda related, but it also looks like onScrollEndDrag might be called twice in the case when you also have momentum when lifting your finger on a drag.

PR #1029, opened 2022-07-14
feat: adding onPress prop to backdrop component

tarikfp:
## Motivation

There might be specific scenarios where some action needs to be executed on pressing backdrop component.

### DIFF
--- a/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
+++ b/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
@@ -29,6 +29,7 @@ const BottomSheetBackdropComponent = ({
   disappearsOnIndex: _providedDisappearsOnIndex,
   enableTouchThrough: _providedEnableTouchThrough,
   pressBehavior = DEFAULT_PRESS_BEHAVIOR,
+  onPress,
   style,
   children,
 }: BottomSheetDefaultBackdropProps) => {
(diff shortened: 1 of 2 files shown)


joeyfigaro:
@tarikpnr doesn't [`pressBehavior`](https://gorhom.github.io/react-native-bottom-sheet/components/bottomsheetbackdrop#pressbehavior) handle `onPress` use cases?

tarikfp:
> @tarikpnr doesn't [`pressBehavior`](https://gorhom.github.io/react-native-bottom-sheet/components/bottomsheetbackdrop#pressbehavior) handle `onPress` use cases?

As far as I can see that it does not support custom on press function. 

https://github.com/gorhom/react-native-bottom-sheet/blob/4a4a9dedef4f83b620a21e1a5670de0e1020a2f6/src/components/bottomSheetBackdrop/types.d.ts#L42

tarikfp:
@gorhom Hi, any thoughts about this pr ?

carlos-cne:
I'm also following this topic.

src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
```diff
@@ -52,6 +53,8 @@ const BottomSheetBackdropComponent = ({
 
   //#region callbacks
   const handleOnPress = useCallback(() => {
+    onPress?.()
```
gorhom: you need to add `onPress` to the `handleOnPress` dependencies list.

gorhom (changes requested):
@tarikpnr thanks for submitting this PR, i left one comment

src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
tarikfp: @gorhom Hey gorhom, ooopss... you are right, I somehow missed it.

gorhom (approved):

PR #1035, opened 2022-07-20
fix: closed bottom sheet snap point

eastroot1590:
closed bottom sheet can't directly snap to 100% because initial animatedNextPosition is 0 and 100%'s nextPosition is 0 too.

## Motivation
this kind of sheets can't open
```
const snapPoint = ['100%'];
const ref = useRef<BottomSheet>(null);

const handlePress = () => {
    // doesn't work
    ref.current?.snapToIndex(0);
};

return (
  <Container>
    <Button onPress={handlePress} />
    <BottomSheet
      ref={ref}
      index={-1}
      snapPoints={snapPoint}
    />
  </Container>
)
```

in `animateToPositionCompleted`, `animatedNextPosition` always set to `Number.NEGATIVE_INFINITY` so this value can be initial value.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -210,7 +210,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       animateOnMount ? -1 : _providedIndex
     );
     const animatedPosition = useSharedValue(INITIAL_POSITION);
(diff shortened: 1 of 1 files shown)


bpolonia (approved):
I'm facing the same issue. 

I just tested it. This fixes it.

Thanks!

gorhom:
thanks @eastroot1590 !

Issue #1068, opened 2022-08-18
[v4] [web-only]: Backdrop prevents click throught


RobertSasak:
# Bug

I struggle to debug following issue. It only occurs on web. Backdrop component is capturing all the pointer events even if it transparent and the bottom sheet is closed/hidden. When I inspect the backdrop `<div aria-label="Bottom Sheet backdrop")>` it has wrongly set _pointer-events_ to _auto_.

Backdrop has two classes 
```css
.r-pointerEvents-12vffkv > * {
    pointer-events: auto;
}
```
and
```css
.r-pointerEvents-12vffkv {
    pointer-events: none !important;
}
```
And the first one is overriding the second one. Backdrop component seems to be correct at https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx

I suspect that this issue is a special quirk from reanimated package.

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.4.3   |
| react-native                    | 0.68.2   |
| expo | ~45.0.4 |
| react-native-reanimated         | ~2.8.0   |
| react-native-gesture-handler    | ~2.2.1   |

## Steps To Reproduce

1. Clone repo and run example
```sh
git clone https://github.com/gorhom/react-native-bottom-sheet
cd examples
cd expo
expo web
```
2. Press _w_ to open web browser
3. Navigate to Backdrop example screen
4. Click on present modal button

Describe what you expected to happen:

1. When the backdrop is closed I expect to be able to click on the screen buttons like _Close_ or _Collapse_. 

## Reproducible sample code

Provided example has this issue https://github.com/gorhom/react-native-bottom-sheet

RobertSasak:
Similar issue on Android https://github.com/gorhom/react-native-bottom-sheet/issues/807

PR #1076, opened 2022-08-26
fix(web): replace setNativeProps with useState

RobertSasak:
## Motivation

This PR address the issue https://github.com/gorhom/react-native-bottom-sheet/issues/1068 which only occurs on web.

Backdrop is preventing any clicks trought. This happens because _pointerEvents_ are never set to actual DOM node. _pointerEvents_ are set with _setNativeProps_ however _setNativeProps_ is not really supported on _react-native-web_. In this simple case _setNativeProps_ can be replace with simple _useState_ and applying _pointerEvents_ directly on element.

It is easy to verify that this fix works. Simply open expo example and navigate to Modal Backdrop or Advance Backdrop. Buttons were not clickable before the fix.

### DIFF
--- a/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
+++ b/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
@@ -1,4 +1,5 @@
-import React, { memo, useCallback, useMemo, useRef } from 'react';
+import React, { memo, useCallback, useMemo, useState } from 'react';
+import { ViewProps } from 'react-native';
 import Animated, {
   interpolate,
   Extrapolate,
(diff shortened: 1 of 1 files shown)


gorhom (approved):

gorhom:
thanks @RobertSasak for submitting this PR

adithyamenu:
This is still not working on web :(

My RNReanimated version is 2.8.0.

RobertSasak:
@adithyamenu You are right. Following PR and release broke it again. More info here.
https://github.com/gorhom/react-native-bottom-sheet/pull/1100#issuecomment-1271335837

In a meanwhile you can use version 4.4.4.

adithyamenu:
Thank you. 
But I am not sure, because I actually monkey patched this PR to the bottom-sheet@4.4.3 that was already installed in my project, and didn't actually update the package to 4.4.4 or 4.4.5.

PR #1099, opened 2022-09-13
chore: introduce bare-fabric example

AndreiCalazans:
## Motivation

This includes a fabric enabled example that can be used to verify the library is properly working with Fabric. 

The example app doesn't fully work with Fabric due to:
1) The usage of `findNodeHandle`
2) react-navigation's Card component using `setNativeProps`
3) Reanimated's V3 RC crashing when using `useReanimatedRef`. 

I have a follow up PR where I remove `findNodeHandle` usage from this library to make it fully compatible with Fabric. 

cc @gorhom


gorhom:
Thanks @AndreiCalazans for submitting this PR.

I don't think we need to have an example for Fabric, or at least a separate example. I will look into extracting the example into another repo in the feature.

AndreiCalazans:
> Thanks @AndreiCalazans for submitting this PR.
> 
> 
> 
> I don't think we need to have an example for Fabric, or at least a separate example. I will look into extracting the example into another repo in the feature.
> 
> 

Sounds good. It made it easier for me to test both in parallel. The example apps are nicely setup, nice job by the way.

only1chi:
Is there any plan to review and merge this PR? 
I think this PR gets bottom-sheet compliant with reanimated v3.

tomekzaw:
> Reanimated's V3 RC crashing when using useReanimatedRef.

Hello! Could you please submit an issue in Reanimated repo so we can address this problem as soon as possible? Thanks in advance!

PR #1102, opened 2022-09-14
fix: prevent not defined error on web

ivannis:
Checking the `__DEV__` global variable in logger.ts will cause an error in React Native Web. Similar to the issue raised in [@gorhom/react-native-portal](https://github.com/gorhom/react-native-portal/issues/29). This is just a copy of AlanSl's [solution](https://github.com/gorhom/react-native-portal/pull/30) for this problem at @gorhom/react-native-portal.


ivannis:
just a comment to keep it open :)

PR #1123, opened 2022-09-30
React18 types

### LINKED ISSUE #1098: [v4] | [v2] React 18 types
# Bug

React Native has been using React 18 since version 0.69. Therefore, it is recommended to also raise the resolution of `@types/react` package to version 18.

With this version bump there were a few breaking changes, which we have already fixed in our repository. Unfortunately, the TypeScript compiler also complains about bugs in this package.

The option `skipLibCheck` has no effect here, because it only affects `.d.ts` files. But this library also publishes the source `.ts` files to NPM, so they are checked by the TypeScript compiler too.

A potential fix would be to raise the version, but this is likely to cause problems for TypeScript users running version 17 of `@types/react`.
An alternative solution would be to no longer publish the source files to NPM.

## Environment info

| Library                      | Version     |
| ---------------------------- | ----------- |
| @gorhom/bottom-sheet         | ^4.4.4      |
| react-native                 | 0.70.0      |
| react-native-reanimated      | ^3.0.0-rc.2 |
| react-native-gesture-handler | ^2.6.0      |

## Steps To Reproduce

1. Set up a new project with the [React Native Template TypeScript](https://github.com/react-native-community/react-native-template-typescript)
2. Make sure that `@types/react` is used in version 18
3. Install `@gorhom/bottom-sheet` and use it somewhere
4. Run `tsc --noEmit`

That the TypeScript compile does not display any errors.

## Reproducible sample code

Doesn't make sense here, since you can't run the TypeScript compiler check in an expo snack.

stropho:
Closes https://github.com/gorhom/react-native-bottom-sheet/issues/1098

## Motivation
- update Typescript React annotations in order to make the library compatible with react 18 types
- additionally, specific version of react types can be used and provided as optional peer dependecies

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,10 +67,20 @@
     "typescript": "^4.2.4"
   },
   "peerDependencies": {
+    "@types/react": "*",
+    "@types/react-native": "*",
     "react": "*",
     "react-native": "*",
     "react-native-gesture-handler": ">=1.10.1",
     "react-native-reanimated": ">=2.2.0"
+  },
+  "peerDependenciesMeta": {
+    "@types/react-native": {
+      "optional": true
+    },
+    "@types/react": {
+      "optional": true
+    }
   },
   "react-native-builder-bob": {
     "source": "src",
--- a/src/components/bottomSheet/types.d.ts
+++ b/src/components/bottomSheet/types.d.ts
@@ -283,9 +283,9 @@ export interface BottomSheetProps
   footerComponent?: React.FC<BottomSheetFooterProps>;
   /**
    * A scrollable node or normal view.
-   * @type React.ReactNode[] | React.ReactNode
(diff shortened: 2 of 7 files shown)


stropho:
yup, this is ready to be reviewed. Not sure what is it waiting for  😉

stropho:
Pretty please @gorhom :) Can anyone have look? There is absolutely no change in the implementation. Just the React types...

stropho:
still waiting :(

stropho:
ping @gorhom 😉

eps1lon:
https://github.com/gorhom/react-native-bottom-sheet/pull/1335 is also required to be compatible with React 18 types and should probably merged regardless since `getScrollResponder` effectively returns `{}` i.e. any non-nullish value.

eps1lon:
bump

PR #1152, opened 2022-10-18
feat: upgrade for reanimated 2.3x

intergalacticspacehighway:
## Motivation
In reanimated 2.3x, the Extrapolate is not exported. Instead we need to use [Extrapolation](https://docs.swmansion.com/react-native-reanimated/docs/api/miscellaneous/interpolate/#extrapolation-type-object--string). This PR just find and replaces `Extrapolate` to `Extrapolation` 😄


intergalacticspacehighway:
merging this might break for 2.2x users though. Not sure when this exactly broke in reanimated. I couldn't find it in changelogs. This looks related - https://github.com/software-mansion/react-native-reanimated/issues/3246

sanduluca (approved):

PR #1163, opened 2022-11-01
Fix android issues with dynamic snap points and open keyboard + `onAnimate` values

paulrostorp:
## Motivation
I was facing an issue on android when using dynamic snap points and the `adjustPan` setting **when content height changed while the keyboard was open**. I think the following video best showcases the problem:

https://user-images.githubusercontent.com/1866649/199200414-1d78a25c-538d-4021-9f27-4c4c9003d52b.mp4

as you can see, the height of the content changes when the validation text appears or disappears, thereby adding or removing about twenty pixels.

I've narrowed down the issue: because of the keyboard height, the appropriate `snapPoint` index is not found and triggers the intermediary collapse of the sheet.

I've tested this on android and ios:
- with and without dynamic snap points setting
- with single and multiple snap points
- with modal/regular bottom sheet
- with bottomsheet scrollview

I have not found this to cause any regressions, but there may be a better solution to this problem.

Edit: 
Upon further investigation, I also found some issues with `onAnimate` being provided with the wrong values (both ios and android), I've fixed that in https://github.com/gorhom/react-native-bottom-sheet/pull/1163/commits/0c7e1532eaac8cd56488ca4c9221c1afdd2714fc, but I'm starting to think my fixes are not tackling the root issue, @gorhom please advise 🙂

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -693,7 +693,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
         /**
          * store next position
          */
-        animatedNextPosition.value = position;
+        animatedNextPosition.value = position + animatedKeyboardHeight.value;
         animatedNextPositionIndex.value =
           animatedSnapPoints.value.indexOf(position + animatedKeyboardHeight.value);
 
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -602,8 +602,9 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
     );
     const handleOnAnimate = useCallback(
       function handleOnAnimate(toPoint: number) {
+        const keyboardOffset = animatedKeyboardState.value == KEYBOARD_STATE.SHOWN ? animatedKeyboardHeight.value : 0;
         const snapPoints = animatedSnapPoints.value;
-        const toIndex = snapPoints.indexOf(toPoint);
+        const toIndex = toPoint == 0 ? INITIAL_SNAP_POINT : snapPoints.indexOf(toPoint + keyboardOffset);
 
         print({
           component: BottomSheet.name,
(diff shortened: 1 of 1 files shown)


paulrostorp:
for any interested, here is a patch:

`@gorhom+bottom-sheet+4.4.5.patch`
```js
diff --git a/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx b/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
index b166440..ce74512 100644
--- a/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
+++ b/node_modules/@gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx
@@ -693,9 +693,9 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
         /**
          * store next position
          */
-        animatedNextPosition.value = position;
+        animatedNextPosition.value = position + animatedKeyboardHeight.value;
         animatedNextPositionIndex.value =
-          animatedSnapPoints.value.indexOf(position);
+          animatedSnapPoints.value.indexOf(position + animatedKeyboardHeight.value);
```

src/components/bottomSheet/BottomSheet.tsx
```diff
@@ -693,9 +694,10 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
         /**
          * store next position
          */
-        animatedNextPosition.value = position;
-        animatedNextPositionIndex.value =
-          animatedSnapPoints.value.indexOf(position);
+        const keyboardOffset = animatedKeyboardState.value == KEYBOARD_STATE.SHOWN ? animatedKeyboardHeight.value : 0;
+
+        animatedNextPosition.value = position + keyboardOffset;
+        animatedNextPositionIndex.value = position == 0 ? INITIAL_SNAP_POINT : animatedSnapPoints.value.indexOf(position + keyboardOffset);
```
paulrostorp, line 700: When the position was `0` (from a fully extended sheet), the index was still specified as `-1`. But that had the implication that the sheet was closed/closing.
I thought to instead set it to `INITIAL_SNAP_POINT` (-999), as that conveys a better meaning, especially in the context of the `onAnimate` callback

paulrostorp:
not stale cc @gorhom

paulrostorp:
not stale cc @gorhom

paulrostorp:
not stale cc @gorhom

paulrostorp:
Not stale

gorhom:
Hi @paulrostorp, thanks for investigating and submitting this pr, could you please provide a reproducible sample code to test out the issue and validate the proposed solution ? thanks

XantreDev:
Probably it's related with this
#1821

idrakimuhamad:
I think it may not be Android only as I'm having this on my app for both platform. It happened when the content size of the sheet changed. In my case, I have a helper text that changed its content depending on the input validation. I think what happened here is its trying to recalculate the sheet size since I've enabled dynamic sizing. The current workaround is to set a fixed height to the content.


https://github.com/gorhom/react-native-bottom-sheet/assets/1400323/e0a19293-d243-462c-876b-5bbfd45de5a9

brentforder:
I'm giving up on dynamic snap points... It's really not supported. I'm replacing it with a BottomSheet per use-case (per set of snap points / per content type) and a "BottomSheetContext" to globally manage the multiple sheets.

PR #1214, opened 2022-12-05
fix: added missing onClose mock

vivianmauer:

## Motivation

When testing `onClose` prop wasn't called


vivianmauer:
Merge pls?

PR #1227, opened 2022-12-12
update navigator example link

swikars1:
I have just updated the old broken link for navigator example in the docs.


swikars1:
dup

PR #1335, opened 2023-03-31
fix(BottomSheetScrollView): Scroll responder types use mixin

eps1lon:
## Motivation

The scroll responders were typed as `ReactNode` which only were usable because `ReactNode` includes `{}` i.e. any non-nullish value. From [React 18 going forward this is no longer the case](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210) so the types became unusable.

I'm not sure if it actually returned a `ScrollResponderMixin` in 0.62. [Earliest (similar) usage of `getScrollResponder: () => ScrollResponderMixin ` was in 0.63](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7be465810096380d994b685fc7dcde874a0ba47c/types/react-native/v0.63/index.d.ts#L6767).

### DIFF
--- a/src/components/bottomSheetScrollable/types.d.ts
+++ b/src/components/bottomSheetScrollable/types.d.ts
@@ -6,7 +6,6 @@ import type {
   RefObject,
 } from 'react';
 import type {
-  ScrollView,
   VirtualizedListProps,
   ScrollViewProps,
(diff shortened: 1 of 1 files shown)


eps1lon:
bump

PR #1333, opened 2023-03-30
Update react-naigation.md

varemenos:
fix: fix a link in the react-navigation integration documentation page

## Motivation

The link for the Navigator Example in https://gorhom.github.io/react-native-bottom-sheet/react-navigation-integration/ is broken.

Also there is a typo for the file name but I didn't change it because I didn't know if that would break any linking.

PR #1339, opened 2023-04-05
Update react-naigation.md, broken link on the website.

Whitelistedd:
fixed the broken url that was redirecting to the wrong page

## Motivation

i was exploring the website and decided to check the react navigation example which then found out that its broken and redirects to the wrong page


suyundukov (approved):
LGTM.

suyundukov:
By the way there is already a fix for it in #1333.

PR #1352, opened 2023-04-23
docs: update present method

jrozbicki:

## Motivation

1. When looking through docs I didn't found any mention of possiblity of passing data from `present` method to `<BottomSheetModal>` component. I've spotted it in TS declaration. 

2. There should also be update to `<BottomSheetModal>` that explains usage of renderProps:

```
<BottomSheetModal>
 {({ data }) => (...)}
</BottomSheetModal>
```

I figured it out from TS declarations as well.

3. Maybe data should be rather typed as `unknown` instead of any? Or even better `BottomSheetModal` type could take `data` type as an argument?

### DIFF
--- a/docs/modal/methods.md
+++ b/docs/modal/methods.md
@@ -41,16 +41,22 @@ const App = () => {
 
 ### `present`
 
-```ts
-type present = (data?: any) => void;
-```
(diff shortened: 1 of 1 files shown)


gorhom:
thanks @jrozbicki for submitting this PR 🙌

PR #1346, opened 2023-04-15
Fix reanimated freezing Keyboard module

janicduplessis:
## Motivation

Reanimated captures and freezes the whole Keyboard module since we reference it from a worklet. To avoid this we can create the function only once at the top of the module from the JS context. This seems to happen only with reanimated v3.

PR #1354, opened 2023-04-23
fix: keyboard dismiss with react-native-reanimated@3

samuelscheit:
react-native-reanimated version 3 [does not support](https://github.com/software-mansion/react-native-reanimated/issues/4177) passing certain objects to worklets.

The following error occurs, when you swipe to dismiss a Bottom Sheet with a BottomSheetTextInput focused:

```js
 LOG  Error: Trying to access property `dismiss` of an object which cannot be sent to the UI runtime.
    at get (/node_modules/react-native-reanimated/src/reanimated2/shareables.ts:69:26)
    at handleOnEnd (/node_modules/@gorhom/bottom-sheet/src/hooks/useGestureEventsHandlersDefault.tsx:299:31)
    at anonymous (/node_modules/@gorhom/bottom-sheet/src/hooks/useGestureHandler.ts:59:20)
    at anonymous (/node_modules/react-native-reanimated/src/reanimated2/hook/useAnimatedGestureHandler.ts:69:21)
    at handleAndFlushAnimationFrame (/node_modules/react-native-reanimated/src/reanimated2/core.ts:123:17)
    at apply (native)
    at callGuardDEV (worklet_1892348396866:3:7)
```

To fix this issue the dismiss function needs to be wrapped that can be sent to the worklet:
```js
const dismissKeyboard = () => Keyboard.dismiss()

...

runOnJS(dismissKeyboard)();
```

### DIFF
--- a/src/hooks/useGestureEventsHandlersDefault.tsx
+++ b/src/hooks/useGestureEventsHandlersDefault.tsx
@@ -225,6 +225,7 @@ export const useGestureEventsHandlersDefault: GestureEventsHandlersHookType =
           animatedScrollableContentOffsetY,
         ]
       );
+    const dismissKeyboard = () => Keyboard.dismiss()
     const handleOnEnd: GestureEventHandlerCallbackType<GestureEventContextType> =
       useWorkletCallback(
         function handleOnEnd(
@@ -296,7 +297,7 @@ export const useGestureEventsHandlersDefault: GestureEventsHandlersHookType =
                 absoluteY > WINDOW_HEIGHT - animatedKeyboardHeight.value
               )
             ) {
-              runOnJS(Keyboard.dismiss)();
+              runOnJS(dismissKeyboard)();
             }
           }
 


malikzype:
Facing the same issue... due to this sometimes the backdrop remains and creates an overlay which prevents the UI to be accessible. @SamuelScheit Did u fix it by changing the runOnJS code?

suyundukov:
Duplicate #1346?

samuelscheit:
> Facing the same issue... due to this sometimes the backdrop remains and creates an overlay which prevents the UI to be accessible. @SamuelScheit Did u fix it by changing the runOnJS code?

I didn't had a problem with the overlay, but with the keyboard not dismissing.
But your problem might also be caused by a similar issue with reanimated V3.

samuelscheit:
@suyundukov you are right, #1346 already fixes the same issue.
The other PR should get merged

Issue #1366, opened 2023-05-11
[v4] ReanimatedError: Tried to synchronously call a non-worklet function on the UI thread. 


beqramo:
# Bug

Hi,
I swipe down the modal to close it and it throws this error.
I think call to the function is wrong

PR has been created: 
https://github.com/gorhom/react-native-bottom-sheet/pull/1367
<img width="439" alt="image" src="https://github.com/gorhom/react-native-bottom-sheet/assets/40486471/5e9d6a67-5a7e-49b8-b581-f2f270e58815">

<img width="446" alt="image" src="https://github.com/gorhom/react-native-bottom-sheet/assets/40486471/6b4175a2-da99-49a7-bfa0-936ff9422544">

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.4.6   |
| react-native                    | 0.71   |
| react-native-reanimated         | ~3.1.0   |
| react-native-gesture-handler    | ~2.9.0   |

## Steps To Reproduce

1.create bottom-sheet input in the bottom-sheet modal
2. try to swipe down to close the modal
3. lib will crash on keyboard close

Describe what you expected to happen:

1. it shouldn't crash on close
2.

## Reproducible sample code

```tsx
const Modal = () =>{
 const ref = useRef(null);
  const sheetRef = useRef(null);



 const renderBackdrop = useCallback(
    (props) => (
      <BottomSheetBackdrop
        {...props}
        disappearsOnIndex={0}
        appearsOnIndex={1}
      />
    ),
    [],
  );
  
return  (
  <BottomSheet
      ref={sheetRef}
      snapPoints={[250]}
      backdropComponent={renderBackdrop}
    >
      <View>
         <BottomSheetTextInput 
           placeholder={'Enter Amount'}
            keyboardType={'decimal-pad'}
            value={value}
            ref={ref}
         />
      </View>
    </BottomSheet>
   )
}
```

starnizar:
have same issue 
there is path of that hook in warning  `/node_modules/@garhom/bottom-sheet/src/hooks/useGestureEventHandlersDefault.tsx`
hope will be useful

[link](https://ibb.co/tMswC77) to image with error

beqramo:
@starnizar you can use the pr  code changes above that I created.

akanoce:
The same happens to me on `4.4.6` + reanimated `2.14.4`
[...]

PR #1367, opened 2023-05-11
fix: crash on swipe down

beqramo:
this PR fixes the issue faced here: 
https://github.com/gorhom/react-native-bottom-sheet/issues/1366

## Motivation

keyboard close was being called incorrectly which was causing a crash on the device.
This fix was tested on device

### DIFF
--- a/src/hooks/useGestureEventsHandlersDefault.tsx
+++ b/src/hooks/useGestureEventsHandlersDefault.tsx
@@ -298,7 +298,7 @@ export const useGestureEventsHandlersDefault: GestureEventsHandlersHookType =
                 absoluteY > WINDOW_HEIGHT - animatedKeyboardHeight.value
               )
             ) {
-              runOnJS(dismissKeyboard)
+              runOnJS(dismissKeyboard)();
             }
           }
 
--- a/src/hooks/useGestureEventsHandlersDefault.tsx
+++ b/src/hooks/useGestureEventsHandlersDefault.tsx
@@ -21,6 +21,7 @@ type GestureEventContextType = {
   isScrollablePositionLocked: boolean;
 };
 
+const dismissKeyboard = Keyboard.dismiss
 
 export const useGestureEventsHandlersDefault: GestureEventsHandlersHookType =
   () => {
@@ -297,7 +298,7 @@ export const useGestureEventsHandlersDefault: GestureEventsHandlersHookType =
                 absoluteY > WINDOW_HEIGHT - animatedKeyboardHeight.value
               )
             ) {
-              runOnJS(Keyboard.dismiss);
+              runOnJS(dismissKeyboard)
             }
           }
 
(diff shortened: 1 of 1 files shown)


Yassdrk (approved):

src/hooks/useGestureEventsHandlersDefault.tsx
```diff
@@ -298,7 +297,7 @@ export const useGestureEventsHandlersDefault: GestureEventsHandlersHookType =
                 absoluteY > WINDOW_HEIGHT - animatedKeyboardHeight.value
               )
             ) {
-              dismissKeyboardOnJs();
+              runOnJS(Keyboard.dismiss);
```
LuisRego: This will crash with `reanimated > 3.x`. My recommendation  -> [94cf11e](https://github.com/gorhom/react-native-bottom-sheet/commit/94cf11eb43f9cc2a356da7e7967ec63baf390a74#diff-30e33da2afcfb6340f6c7eff36ac5932d6cc5e6f7f5e895d5bb4bd7e48e6d984R301)

beqramo: Hi,
I haven't tested on V3.
So it means we need to bump the required reanimated version on this lib?
Or what is your suggestion? 
That way it crashes on `reanimated < 3.x`

Thanks

beqramo: Update:
I tested on   ` "react-native-reanimated": "~3.1.0"` and it works perfectly

LuisRego: Just tested in `"react-native-reanimated": "3.1.0"`.

<div>

</div>

beqramo: yes exactly, that is a fix that this PR has.
That was the cause of the issue.

MaxAst: Tried this, but got this error now `ReanimatedError: Trying to access property `dismiss` of an object which cannot be sent to the UI runtime., js engine: reanimated`

vanenshi: > Tried this, but got this error now `ReanimatedError: Trying to access property `dismiss` of an object which cannot be sent to the UI runtime., js engine: reanimated`

Same


</div>

vanenshi: something like this works without problem  (@LuisRego)

```tsx
const dismissKeyboard = () => {
  Keyboard.dismiss()
}

...

runOnJS(dismissKeyboard)();
```

beqramo:
Hi Guys, 
I'm on reanimated ~3.1.0.

Update/Edit:
I was able to run 3.2.0.
Just noticed that on a patch file locally, I had a different way of the fix :).
I'm pushing the correct variant.

Please take a look
Thanks

src/hooks/useGestureEventsHandlersDefault.tsx
beqramo: https://github.com/gorhom/react-native-bottom-sheet/pull/1367#issuecomment-1574224496

gorhom (approved):

gorhom:
thanks @beqramo for submitting this CR , i will prepare a new patch release today with this fix <3

Issue #1374, opened 2023-05-17
[v4] Sheet re-opens after closing on Android when keyboard present - only with 100% detent.


pgill-rbi:
# Bug

On Android specifically, the Sheet would intermittently re-open after closing it in the following scenario:

Keyboard is open
Detent is 100%
This was an issue within the @gorhom/bottom-sheet library where the function normalizeSnapPoints would not change the true pixel value of a 100% detent. It always equates to 0. This is specific to Android with adjustResize as the keyboard behaviour.

A condition was added as a patch to prevent exiting a function (onSnapPointsChange) which allows the Sheet to react to snap point changes mid-animation.

***A PR to solve this has been submitted here:*** https://github.com/gorhom/react-native-bottom-sheet/pull/1375

https://github.com/gorhom/react-native-bottom-sheet/assets/105675561/ad83c494-5d41-49cd-87d3-913ba115f8f4

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            |4.4.4   |
| react-native                    | 0.69.6   |
| react-native-reanimated         | 2.10.0  |
| react-native-gesture-handler    | ~2.5.0   |

## Steps To Reproduce

1. On Android, Open bottomsheet with a bottomsheetflatlist inside, set only one detent at 100%.
2. Drag down Sheet while keyboard is open ( to close it)

Describe what you expected to happen:

1. BottomSheet should close.

What actually happens:

1. BottomSheet closes and reopens.

## Reproducible sample code

```
 const styles = StyleSheet.create({
    container: {
      flex: 1,
      paddingTop: 200,
    },
    contentContainer: {
      backgroundColor: "white",
    },
    itemContainer: {
      padding: 6,
      margin: 6,
      backgroundColor: "#eee",
    },
  });
  const handleSheetChange = useCallback((index) => {
    console.log("handleSheetChange", index);
  }, []);
  const handleClosePress = useCallback(() => {
    sheetRef.current?.close();
  }, []);
  const data = useMemo(
    () =>
      Array(50)
        .fill(0)
        .map((_, index) => `index-${index}`),
    []
  );
  const renderItem = useCallback(
    ({ item }) => (
      <View style={styles.itemContainer}>
        <Text>{item}</Text>
      </View>
    ),
    []
  );
  return (
    <GestureHandlerRootView style={{ flex: 1 }}>
      <Box flex={1} alignItems="center" mt="6">
        <Button
[...]

PR #1375, opened 2023-05-17
fix: Case to prevent Android sheet from reopening using 100% detent

pgill-rbi:
Please see full detail of the bug addressed here: https://github.com/gorhom/react-native-bottom-sheet/issues/1374

## Motivation

On Android specifically, the Sheet would intermittently re-open after closing it in the following scenario:

Keyboard is open
Detent is 100%
This was an issue within the @gorhom/bottom-sheet library where the function normalizeSnapPoints would not change the true pixel value of a 100% detent. It always equates to 0. This is specific to Android with adjustResize as the keyboard behaviour.

A condition was added as a patch to prevent exiting a function (onSnapPointsChange) which allows the Sheet to react to snap point changes mid-animation.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -1284,15 +1284,28 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
     useAnimatedReaction(
       () => ({
         snapPoints: animatedSnapPoints.value,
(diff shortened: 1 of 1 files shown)


gorhom:
thanks @pgill-rbi for submitting this CR, i have pushed another one to address this issue with a scalable approach
https://github.com/gorhom/react-native-bottom-sheet/pull/1392

pgill-rbi:
@gorhom the problem still persists with #1392 using release 4.4.7.

PR #1394, opened 2023-06-07
mock BottomSheet list with ReactNative list equivalent

gkueny:
## Motivation

In react-native, if we run test on component where we use `BottomSheetScrollView` / `BottomSheetSectionList` / `BottomSheetFlatList` `BottomSheetVirtualizedList` with `@gorhom/bottom-sheet/mock` mock, It doesn't work. The content of these components is not rendered. 

That's because these components are replaced by `BottomSheetComponent` which does not implement `renderItem` props.

## Solution

Using react-native components is equivalent to solving the above-mentioned problem.

### DIFF
--- a/mock.js
+++ b/mock.js
@@ -9,6 +9,7 @@
  */
 
 const React = require('react');
+const ReactNative = require('react-native');
 
 const NOOP = () => {};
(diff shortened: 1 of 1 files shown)


gorhom:
thanks for submitting this CR

PR #1412, opened 2023-06-23
chore: updated react native and other deps

gorhom:
## Motivation

Updated React Native to 0.72 and other deps to prepare for the next release


RichardLindhout:
@gorhom can you release this on npm?

PR #1420, opened 2023-06-26
fix: keyboard on focus

### LINKED ISSUE #1328: [v4] On Input focus Modal goes behind keyboard on iOS and Android
# Bug

Hi,
I have a screen where by default modal is opened on screen open/focus I'm triggering Textinput Focus, and sometimes modal goes behind the keyboard like it is shown below. I tried time out on input focus, and it helped but not fully, sometimes it is still happening. From what we saw it is only happening on ios. 

## Environment info

| @gorhom/bottom-sheet            | 4.4.5   |
| react-native                    | 0.68   |
| react-native-reanimated         | 2.9.1   |
| react-native-gesture-handler    | 2.9.0  |

## Steps To Reproduce

1. add Input in the Modal
2. Trigger Input focus it on screen focus
3. Sometimes you may see that modal goes behind the keyboard.

1.
2.

## Reproducible sample code

I simplified the component for this example.

```tsx
const Modal = () =>{
 const ref = useRef(null);
  const sheetRef = useRef(null);

  useEffect(() => {
    setTimeout(() => {
      ref.current?.focus();
    }, 800);
  }, []);


    (props) => (
      <BottomSheetBackdrop
        {...props}
        disappearsOnIndex={-1}
        appearsOnIndex={1}
        pressBehavior={ 'none'}
        enableTouchThrough={false}
      />
    ),
    [],
  );
  
return  (
  <BottomSheet
      ref={sheetRef}
      snapPoints={[250]}
      enablePanDownToClose={false}
      backdropComponent={renderBackdrop}
      enableContentPanningGesture={false}
      backgroundStyle={{ flex: 1 }}
      style={{
        borderRadius: 24,
      }}
      keyboardBehavior={undefined} // I have tried all other options too but still not stable
      keyboardBlurBehavior="none"
      android_keyboardInputMode="adjustPan"
      handleComponent={null}
      handleHeight={0}
      bottomInset={0}
    >
      <View>
         <BottomSheetTextInput 
           placeholder={'Enter Amount'}
            keyboardType={'decimal-pad'}
            value={value}
            onFocus={onFocusLocal}
            onBlur={onBlurLocal}
            ref={ref}
         />
      </View>
    </BottomSheet>
   )
}
```

Thank you!

beqramo:
Fixes: #1328 

## Motivation
Sometimes, especially when the view is newly created `measure` function return undefined on `PageY` that was causing hiding modal behind the keyboard as  `animatedContainerOffset.value.bottom` was `NaN` in `BottomSheet.tsx` file
image

With this change, I'm passing the default value 0 so that it will not have the bug.

With the current change, everything is working fine on both platforms.

Thanks

### DIFF
--- a/src/components/bottomSheetContainer/BottomSheetContainer.tsx
+++ b/src/components/bottomSheetContainer/BottomSheetContainer.tsx
@@ -54,7 +54,7 @@ function BottomSheetContainerComponent({
             right: 0,
             bottom: Math.max(
               0,
-              WINDOW_HEIGHT - (pageY + height + (StatusBar.currentHeight ?? 0))
+              WINDOW_HEIGHT - ((pageY ?? 0) + height + (StatusBar.currentHeight ?? 0))
             ),
           };
         }
--- a/src/components/bottomSheetContainer/BottomSheetContainer.tsx
+++ b/src/components/bottomSheetContainer/BottomSheetContainer.tsx
@@ -49,7 +49,7 @@ function BottomSheetContainerComponent({
       containerRef.current?.measure(
         (_x, _y, _width, _height, _pageX, pageY) => {
           containerOffset.value = {
-            top: pageY,
+            top: pageY ?? 0,
             left: 0,
             right: 0,
             bottom: Math.max(


gorhom:
thanks @beqramo!

CarlMenke:
Any idea when this fix will be live? I don't see it in 4.4.7. Any estimation would be very helpful, thanks!

beqramo:
Hi,
It is fixed in version 5 alpha

abdulragib:
> Hi, It is fixed in version 5 alpha

tell me , which package , i am confused becuase i installed lots of package

brentforder:
> Hi, It is fixed in version 5 alpha

@gorhom / @beqramo 
The fix is not working for me with the new v5 alpha (5.0.0-alpha.11).
I see that this build includes the changes per #1420
The iOS keyboard is still covering my inputs when popping up on focus, as described in #1328

brentforder:
> > Hi, It is fixed in version 5 alpha
> 
> @gorhom / @beqramo The fix is not working for me with the new v5 alpha (5.0.0-alpha.11). I see that this build includes the changes per #1420 The iOS keyboard is still covering my inputs when popping up on focus, as described in #1328

Update: False alarm., sorry. It's working now that I switched from React Native Paper's TextInput to BottomSheetTextInput. To take control of this virtual keyboard behavior seems to require BottomSheetTextInput.

PR #1425, opened 2023-06-28
fix: always pass ref object to mountSheet

NoahCardoza:
Motivation
----------

This pull request addresses an issue encountered while using the `useBottomSheetModal` hook in conjunction with a wrapper around the bottom sheet component. The wrapper relies on the `forwardRef` functionality and requires access to the bottom sheet's ref. However, a problem arises when utilizing the `BottomSheetModalProvider`, specifically in the `handleMountSheet` function. This function pushes new sheets to an array that tracks all open sheets, which can then be utilized by the `handleDismiss` and `handleDismissAll` methods. The issue stems from the assumption that the refs stored in the objects within the `sheetsQueueRef` are always objects.

Upon investigation, it became apparent that invoking `dismiss` or `dismissAll` through the `useBottomSheetModal` hook had no effect within our app. A closer examination revealed that the line `sheetToBeDismissed.ref?.current?.dismiss();` failed silently due to the presence of `?.current` since our `sheetToBeDismissed.ref` resolved to a function instead of a `Ref` object.

Solution
--------

This pull request resolves the aforementioned issue by encapsulating the incoming ref from `forwardRef` and creating a callback function that is passed to `useImperativeHandle`. This approach ensures that the forward `ref` and the new `internalRef` remain synchronized. Consequently, the `internalRef` is then used in `mountSheet`, guaranteeing that a ref object is consistently stored within the `sheetsQueueRef` array.

### DIFF
--- a/src/components/bottomSheetModal/BottomSheetModal.tsx
+++ b/src/components/bottomSheetModal/BottomSheetModal.tsx
@@ -10,7 +10,7 @@ import React, {
 import { Portal, usePortal } from '@gorhom/portal';
 import BottomSheet from '../bottomSheet';
 import { useBottomSheetModalInternal } from '../../hooks';
-import { print } from '../../utilities';
+import { print, isFunction } from '../../utilities';
 import {
   DEFAULT_STACK_BEHAVIOR,
   DEFAULT_ENABLE_DISMISS_ON_CLOSE,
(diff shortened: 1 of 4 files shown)


gorhom:
Hi @NoahCardoza , could you provide a repo sample code of the issue, i would like to test it out before merging the PR

NoahCardoza:
Sorry, I just saw this. I'm no longer interning for the company that needed this and am too busy with school to create a sample at the moment. They are directly installing `react-native-bottom-sheet` from my fork so merging is not urgent. If I do find time in the future, I may reopen this if I create a sample.

PR #1427, opened 2023-07-01
fix: invalid prop type

IggsGrey:
The `fillParent` prop type of the BottomSheet component is typed as `fullScreen` in the docs, which is inconsistent with the existing code.

This PR fixes the 'typo'


gorhom:
thanks @IggsGrey 🙌

Issue #1416, opened 2023-06-24
[v4] screen resize events triggering animatedIndex change


jspizziri:
# Bug

When a resize event occurs such as a screen rotation (in native) or a window resize (in web) the bottom sheet index can change suddenly. This seems to be due to [`animatedContainerHeight` being used in the calculation of `animatedIndex`](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L466).

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            |  4.4.7   |
| react-native                    | 0.71.6   |
| react-native-reanimated         |  ^2.6.0   |
| react-native-gesture-handler    | ^2.4.2   |

## Steps To Reproduce

**Web**

1. Open the bottom sheet.
2. Resize the window.

**Mobile**
I haven't tested this on mobile but theoretically it should happen there too.

1. Open the bottom sheet.
2. Rotate the device (triggering a change in the screen dimensions)

Describe what you expected to happen:

A window/screen resize should not result in a change in the `animatedIndex`

## Reproducible sample code

N/A

## Notes

I've done a little exploration on fixing this but I'm not super happy with any of my approaches or results. I've effectively tried to set a property in the component [when a resize reaction takes place](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L1360) and then adding a conditional to check if a resize event is the responsible for triggering [a recalc on the `animatedIndex` property](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L502).

If I could get some direction from someone who knows more about the library and the best path forward I'd happily do the implementation and submit a PR!

gorhom:
i will look into this shortly

gorhom:
I tried with rotating the device, and it seems working as expected 

https://github.com/gorhom/react-native-bottom-sheet/assets/4061838/d74c6efd-62f1-4108-859d-78b608a1cb2e

gorhom:
tested on `v5` web, and it works as expected too.

https://github.com/gorhom/react-native-bottom-sheet/assets/4061838/47b80d29-28e7-4054-a189-6ae72ec82ed9

jspizziri:
[...]

PR #1435, opened 2023-07-07
fix(bottomsheet): fix window collapse resulting in an update in the a…

jspizziri:
…nimatedIndex of the bottomsheet

fix https://github.com/gorhom/react-native-bottom-sheet/issues/1416

## Motivation

As documented in the above issue a window collapse will cause the bottomsheet to recalculate its `animatedIndex` in certain scenarios. If the window is decreasing in size the user doesn't expect a change or closure of the bottom sheet index.

This behavior can be observed in the following scenarios:

1. Have a bottom sheet position sufficiently close to the bottom of the screen (lets say `100px`).
1. On web, with the sheet open to the smallest index, minimize or resize the browser window (this must be a quick action and a sufficient size differential).
1. On native Android with `android:windowSoftInputMode="adjustResize"` and the sheet open to the smallest index, trigger the keyboard from an input that is _NOT_  inside the sheet, but is outside of it (e.g. a search text box).

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -388,6 +388,26 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       return SCROLLABLE_STATE.LOCKED;
     });
     // dynamic
+    const isContainerHeightCollapsing = useSharedValue<boolean>(false);
+    useAnimatedReaction(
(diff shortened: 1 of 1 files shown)


jspizziri:
@gorhom would love your feedback on this one. It doesn't feel super elegant, but it does seem to be getting the job done for me. I imagine someone with more knowledge of the lib might be able to identify something more elegant.

PR #1402, opened 2023-06-14
Drawer scrollable with dynamic max height 

ororsatti:
Adding a hook to have a drawer with a max height that will dynamically change its height until reaching the limit, following this discussion of mine:  https://github.com/gorhom/react-native-bottom-sheet/discussions/1363
## Motivation
People are asking for this feature for a while now, and asked me in the discussion to share the code.
Closes: (https://github.com/gorhom/react-native-bottom-sheet/issues/658 , https://github.com/gorhom/react-native-bottom-sheet/issues/32)
### Example: 

https://github.com/gorhom/react-native-bottom-sheet/assets/109456266/ecb62d93-6c9e-4122-8b4a-7b3a1df978f0

Issue #658, opened 2021-09-23
[v4] Dynamic height with FlatList not scrolling.


alexisloiselle:
# Bug

When using the `useBottomSheetDynamicSnapPoints` together with `BottomSheetFlatList`, it's not possible to scroll the list anymore.

https://user-images.githubusercontent.com/16881226/134593038-28bd8fc6-841c-4e91-a380-b0a231b7e70d.MP4

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.0.3   |
| react-native                    | 0.64.2   |
| react-native-reanimated         | 2.2.2   |
| react-native-gesture-handler    | 1.10.3   |

## Steps To Reproduce

1. add a `BottomSheet` with the `animatedSnapPoints`, `animatedContentHeight` and `animatedHandleHeight` provided by `useBottomSheetDynamicSnapPoints`
2. set the `handleContentLayout` to the child
3. add a big enough list to overflow the screen

Describe what you expected to happen:

1. it's not possible to scroll anymore

## Reproducible sample code

Here's a Snack to reproduce:
https://snack.expo.dev/@alexisloiselle/dynamic-height-with-flatlist-not-scrolling

VyacheslavAA:
Hello @alexisloiselle !

I had the same issue with `BottomSheetScrollView`.

As a temporary solution you can use `BottomSheetScrollView` and change `height` property for `BottomSheetScrollView` using `useAnimatedStyle` to constraint height of `BottomSheetScrollView` .
Also, to prevent infinity loop for `handleContentLayout` you need to add additional `View` with onLayout handler as `BottomSheetScrollView` child.

Here is example:

```
import BottomSheet, {
  BottomSheetScrollView,
  useBottomSheetDynamicSnapPoints
} from '@gorhom/bottom-sheet';
import React, { useMemo } from 'react';
import { useWindowDimensions, View } from 'react-native';
import { useAnimatedStyle } from 'react-native-reanimated';

export const YourComponent = () => {
  const { height: windowHeight } = useWindowDimensions();

  const initialSnapPoints = useMemo(() => [350, 'CONTENT_HEIGHT'], []);
  const {
    animatedHandleHeight,
    animatedSnapPoints,
    animatedContentHeight,
    handleContentLayout,
  } = useBottomSheetDynamicSnapPoints(initialSnapPoints);

[...]

PR #1510, opened 2023-09-08
Eli nathan/dynamic size bottom sheet with max height

Eli-Nathan:
Inspired by https://github.com/gorhom/react-native-bottom-sheet/pull/1402 @ororsatti
Adding a hook to have a sheet with a max height that will dynamically change its height until reaching the limit, following this discussion: https://github.com/gorhom/react-native-bottom-sheet/discussions/1363

- Added `useBottomSheetMaxHeight` hook
- Call above hook from `useBottomSheetDynamicSnapPoints`
- Updated the example to show the max height & scrolling functionality.
  - Example update also changes it from adding blank empty height to adding new rows of content 

Motivation
People are asking for this feature for a while now, and asked me in the discussion to share the code.
Closes: (https://github.com/gorhom/react-native-bottom-sheet/issues/658 , https://github.com/gorhom/react-native-bottom-sheet/issues/32)

## Example
<video src="https://github.com/gorhom/react-native-bottom-sheet/assets/32842898/dc0e9648-e679-48dd-8951-b9dc7009751e" />

### DIFF
--- a/src/hooks/useBottomSheetMaxHeight.ts
+++ b/src/hooks/useBottomSheetMaxHeight.ts
@@ -1,0 +1,39 @@
+import { useWindowDimensions } from 'react-native';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
+
+type MaxHeight = `${string}%` | number;
+
+const getPercentageMaxHeight = (
+  maxHeight: `${string}%`,
+  safeMaxheightValue: number
+) => {
+  return safeMaxheightValue * (parseFloat(maxHeight) / 100);
+};
+
+const getNumberMaxHeight = (maxHeight: number, safeMaxheightValue: number) => {
+  if (maxHeight > safeMaxheightValue) {
+    return safeMaxheightValue;
+  }
+  return maxHeight;
+};
+
+const useBottomSheetMaxHeight = (maxHeight?: MaxHeight) => {
+  const { height: deviceHeight } = useWindowDimensions();
+  const { top: topSafeAreaInset } = useSafeAreaInsets();
+  if (!maxHeight) {
+    return undefined;
+  }
+  const safeDeviceHeight = deviceHeight - topSafeAreaInset;
+  const maxHeightIsPercentage = typeof maxHeight === 'string';
+  const maxHeightNumber = maxHeightIsPercentage
+    ? getPercentageMaxHeight(maxHeight as `${string}%`, safeDeviceHeight)
+    : getNumberMaxHeight(maxHeight as number, safeDeviceHeight);
+
+  const childViewStyle = {
+    maxHeight: maxHeightNumber,
+  };
+
+  return childViewStyle;
+};
+
+export default useBottomSheetMaxHeight;
--- a/example/app/src/screens/advanced/DynamicSnapPointExample.tsx
+++ b/example/app/src/screens/advanced/DynamicSnapPointExample.tsx
@@ -20,7 +20,8 @@ const DynamicSnapPointExample = () => {
     animatedSnapPoints,
     animatedContentHeight,
     handleContentLayout,
-  } = useBottomSheetDynamicSnapPoints(initialSnapPoints);
+    childViewMaxHeightStyle,
+  } = useBottomSheetDynamicSnapPoints(initialSnapPoints, '90%');
 
   // callbacks
   const handleIncreaseContentPress = useCallback(() => {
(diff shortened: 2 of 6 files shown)


ororsatti:
May the odds be in your favour !

src/hooks/useBottomSheetMaxHeight.ts
```diff
@@ -0,0 +1,39 @@
+import { useWindowDimensions } from 'react-native';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
```
gorhom, line 2: could we use internal container height instead using `react-native-safe-area-context` library ?

gorhom:
first, thanks @ororsatti for submitting the initial PR, i do apologies for not reviewing it on time , your contribution is appreciated.

second, @Eli-Nathan i left a comment about the usage of `react-native-safe-area-context`, im wondering if we could just use the internal container height value ?

Eli-Nathan:
Thanks so much!
My thinking was that bottom sheets (especially modals) can wrap the entire app including the react navigation elements like the header and bottom tabs. Using the safe area guarantees that on newer iOS devices that the bottom sheet never goes higher than the notch and covers the status bar.
Also, actually getting the parent container height would need to be done by the developer using it and passed in to this hook as the library doesn't have context of where it's rendered right? Felt like it would make user error more likely. What do you think? You'll know this code far better than myself.

Eli-Nathan:
@gorhom examples with and without safe area when modal provider wraps the whole app 😄 

| With safe area | Without safe area |
| ---------------- | ---------------- |
|  |  |

gorhom:
I have already expose the container height [`animatedContainerHeight`](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/contexts/internal.ts#L51C3-L51C26) from [`useBottomSheetInternal`](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/hooks/useBottomSheetInternal.ts), which you can access with unsafe flag, since the hook is out of the bottom sheet node

```tsx

const { animatedContainerHeight } = useBottomSheetInternal(true);

```

Eli-Nathan:
It's telling me it's not available 😞 

Screenshot 2023-09-08 at 19 54 38

gorhom:
hmmm ,, let me think of a different approach .. my main concern is now we are adding another hard dependency to the project which is something i try to avoid

gorhom:
i was thinking about this for a while now, what if we moved the while dynamic sizing internally , where users do not need to do anything 🧐 , except using a view from the library ( scrollable or just static view )

Eli-Nathan:
Edit: Ah wait it works without the boolean...
Although technically we wouldn't be using it from within the bottom sheet so the internal context throws an error

Eli-Nathan:
> where users do not need to do anything

Hmm, would that still allow them to specify a maxHeight though?

gorhom:
we could introduce a prop for max snap point to address that issue, so users who wants dynamic size they will need to provide two props:

```
<BottomSheet enableDynamicSizing={true} maxSnapPoint={500 | "100%"} />
```
with this, we will remove the dynamic hook and remove handling layout complexity

Eli-Nathan:
So we'd be applying these styles to `<BottomSheetContainer />` within the main bottom sheet component based on the presence of a enableDynamicSizing prop?

I've had a look at that component and it's very complex.

What are your thoughts on doing applying this as a solution for V4 and the enableDynamicSizing solution for V5 since it seems like it'd be a much bigger refactor?

gorhom:
>So we'd be applying these styles to <BottomSheetContainer /> within the main bottom sheet component based on the presence of a enableDynamicSizing prop?

no, i would apply it to the content container and scrollables 

 >What are your thoughts on doing applying this as a solution for V4 and the enableDynamicSizing solution for V5 since it seems like it'd be a much bigger refactor?
 
 mm i think the code should be sharable with v5 if we do it correctly for v4. im going to draft something as a poc and get back to you

Eli-Nathan:
Sounds good! Thanks again!

ororsatti:
> i was thinking about this for a while now, what if we moved the while dynamic sizing internally , where users do not need to do anything 🧐 , except using a view from the library ( scrollable or just static view ) 

First of all, I'm so glad to hear from you, I thought this lib is going to dei down slowly,
Second, It seems like a big project and I'll be more than happy to help move it, and v5.
I'd be happy to contact you & @Eli-Nathan and maybe get a quick rundown on your lib, and start helping you with it.

Eli-Nathan:
@ororsatti Count me in, this has been hugely valuable to me in multiple projects so happy to help where I can

gorhom:
@Eli-Nathan @ororsatti managed to get dynamic sizing by prop working with View with the approach i suggested , going to apply it on scrollables and i'll prepare a draft PR by tomorrow

gorhom:
closing this PR in favour of #1513, which it has been merged. Thanks @Eli-Nathan

PR #1512, opened 2023-09-09
fix: use correct link, grammar

Rexogamer:
## Motivation

The current URL leads to a 404 - it seems they moved the installation instructions to a different page.


gorhom:
thanks @Rexogamer for submitting this PR 🚀

PR #1529, opened 2023-09-15
fix: `dismissAll` not dismissing all bottom sheets

david-gomes5:
## Motivation

Fixes a regression introduced in v4.4.8 from https://github.com/gorhom/react-native-bottom-sheet/pull/1309 that breaks `dismissAll` from the `useBottomSheetModal` hook when stacking multiple bottom sheet modals.

## Reproduction

- Run the example app
- Press `Stack Modals`
- Press multiple `Present Modal [A|B|C]` buttons to stack multiple bottom sheets
- Press `Dismiss All Modal` button

Fails to dismiss all bottom sheets

### DIFF
--- a/src/components/bottomSheetModal/BottomSheetModal.tsx
+++ b/src/components/bottomSheetModal/BottomSheetModal.tsx
@@ -376,7 +376,7 @@ const BottomSheetModalComponent = forwardRef<
       handleOnUpdate={handlePortalRender}
       handleOnUnmount={handlePortalOnUnmount}
     >
-      <ContainerComponent>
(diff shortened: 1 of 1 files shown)


gorhom:
good catch @david-gomes5 ! thanks for the PR

bryanprimus:
just about to file an issue about this. calling `dismiss` in sheet1 and `present` in `sheet2` at the same time causing glitch

but seems already fixed by this PR

https://github.com/gorhom/react-native-bottom-sheet/assets/46083126/ad05fbbe-0f16-4158-8b81-f7d24d0a2564

Thanks

PR #1597, opened 2023-10-26
fix: use absolute positioning for BottomSheetContainer in web (v5)

viljark:
## Motivation
With fixed positioning in web, the non-modal bottom sheet is not getting along with [bottom tab navigator](https://reactnavigation.org/docs/bottom-tab-navigator/) and is displayed under it. 
Using `absolute` allows BottomSheetContainer to be positioned relatively to containing elements instead of the viewport.

In native iOS/Android it is working as expected with bottom tab navigator.

Before
image

After 
image

### DIFF
--- a/src/components/bottomSheetContainer/styles.web.ts
+++ b/src/components/bottomSheetContainer/styles.web.ts
@@ -2,8 +2,7 @@ import { StyleSheet } from 'react-native';
 
 export const styles = StyleSheet.create({
   container: {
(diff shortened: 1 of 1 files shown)


viljark:
Not stale

viljark:
Not stale

gorhom:
thanks @viljark for submitting this PR, it will be published on the next release

PR #1623, opened 2023-11-08
Fix `snapToPosition` function call, right away after the mount.

### LINKED ISSUE #1294: [v4] | [v2] snapToPosition opening bottomsheet to maximum screen height and snapToIndex not working
# Bug

snapToPosition method is opening bottomsheet in  maximum screen height and snapToIndex is not working when used in useEffect

## Environment info

| @gorhom/bottom-sheet            | ^4   |
| react-native                    | 0.70.5  |
| react-native-reanimated         | 2.13.0  |
| react-native-gesture-handler    | 2.6.1   |

## Steps To Reproduce

1. create react native app
2. use bottomsheet from library
3. use snapToIndex and snapToPosition inside useeffect or useLayoutEffect

1. open bottomsheet with correct index provided
2. open bottomsheet with correct prosition provided

## Reproducible sample code

````ts
const snapPoints = React.useMemo(
    () => ['10%', '25%', '45%', '65%', '95%'],
    [],
  );

React.useEffect(() => {
      sheetRef.current?.snapToPosition(400); // this opens bottomsheet to maximum screen height
 // or this 
// sheetRef.current?.snapToIndex(3); // this is not working at all
  }, [sheetRef.current , loaded]);


````

### PR DESCRIPTION
**Pull request by beqramo**
Fixes: https://github.com/gorhom/react-native-bottom-sheet/issues/1294

## Motivation
When `snapToPosition` gets called right away after a mount, `animatedContainerHeight` has an initial value of -999, if the prop didn't get passed. It is happening as the package can't calculate height right away on the mount. in simple terms  `onLayout` function takes time to be called and to update the value of `animatedContainerHeight`

That means that `snapToPosition` causes some problems.
To fix the issue, we need to wait for the update of the value of  `animatedContainerHeight` and then continue calculating the `nextPosition` value from `normalizeSnapPoint` function.

Because of that, I added a listener if I see that `animatedContainerHeight` still has an initial value. after I receive the update, I'm following the same process as it was before.

I was forced to update the Reanimated package due to the recent addition of the listener feature. I've been using the updated Reanimated package in my production app for a while now, and it's been working seamlessly, so it shouldn't be a problem.
 
 Thank you.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "react-native": "^0.62.2",
     "react-native-builder-bob": "^0.18.1",
     "react-native-gesture-handler": "^1.10.3",
-    "react-native-reanimated": "^2.8.0",
+    "react-native-reanimated": "^3.5.4",
     "release-it": "^16.1.5",
     "typescript": "^4.2.4"
   },
@@ -72,7 +72,7 @@
     "react": "*",
     "react-native": "*",
     "react-native-gesture-handler": ">=1.10.1",
-    "react-native-reanimated": ">=2.2.0"
+    "react-native-reanimated": ">=3.0.0"
   },
   "peerDependenciesMeta": {
     "@types/react-native": {
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -807,39 +807,58 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
           },
         });
 
-        /**
-         * normalized provided position.
-         */
-        const nextPosition = normalizeSnapPoint(
-          position,
-          animatedContainerHeight.value
-        );
+        const snapToPosition = () => {
+          /**
+           * normalized provided position.
+           */
+          const nextPosition = normalizeSnapPoint(
+            position,
+            animatedContainerHeight.value
+          );
+
+          /**
+           * exit method if :
+           * - layout is not calculated.
+           * - already animating to next position.
+           * - sheet is forced closing.
+           */
+          if (
+            !isLayoutCalculated ||
+            nextPosition === animatedNextPosition.value ||
+            isForcedClosing.value
+          ) {
+            return;
+          }
+
+          /**
+           * mark the new position as temporary.
+           */
+          isInTemporaryPosition.value = true;
+
+          runOnUI(animateToPosition)(
+            nextPosition,
+            ANIMATION_SOURCE.USER,
+            0,
+            animationConfigs
+          );
+        };
 
         /**
-         * exit method if :
-         * - layout is not calculated.
-         * - already animating to next position.
-         * - sheet is forced closing.
+         * Checks whether the library has measured the container's height or if it was passed from props.
+         * If the height is not measured, the function waits for the value to be updated.
+         * The variable 'isContainerHeightCalculated' cannot be solely trusted,
+         * as it only checks if the value is neither undefined nor null.
+         * Fixes: https://github.com/gorhom/react-native-bottom-sheet/issues/1294
          */
-        if (
-          !isLayoutCalculated ||
-          nextPosition === animatedNextPosition.value ||
-          isForcedClosing.value
-        ) {
-          return;
-        }
-
-        /**
-         * mark the new position as temporary.
-         */
-        isInTemporaryPosition.value = true;
-
-        runOnUI(animateToPosition)(
-          nextPosition,
-          ANIMATION_SOURCE.USER,
-          0,
-          animationConfigs
-        );
+        if (animatedContainerHeight.value === INITIAL_CONTAINER_HEIGHT)
+          runOnUI(() => {
+            const id = Math.random();
+            animatedContainerHeight.addListener(id, () => {
+              runOnJS(snapToPosition)();
+              animatedContainerHeight.removeListener(id);
+            });
+          })();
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -62,7 +62,7 @@
     "react-native": "^0.62.2",
     "react-native-builder-bob": "^0.18.1",
     "react-native-gesture-handler": "^1.10.3",
-    "react-native-reanimated": "^2.8.0",
+    "react-native-reanimated": "^3.5.4",
```
gorhom, line 65: why updating reanimated ?

src/components/bottomSheet/BottomSheet.tsx
```diff
@@ -807,39 +807,58 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
...
+            nextPosition,
+            ANIMATION_SOURCE.USER,
+            0,
+            animationConfigs
+          );
+        };
 
         /**
-         * mark the new position as temporary.
+         * Checks whether the library has measured the container's height or if it was passed from props.
```
gorhom, line 847: why not early exit this method , and ensure to listen to `isLayoutCalculated` ?

gorhom:
thanks @beqramo for submitting this PR, but i think it is not about layout not being ready, `isLayoutCalculated` is responsible to trigger the on mount animation, which waits till container height to be calculated before it become truthful

package.json
gorhom: >I was forced to update the Reanimated package due to the recent addition of the listener feature. I've been using the updated Reanimated package in my production app for a while now, and it's been working seamlessly, so it shouldn't be a problem.

just saw this

gorhom:
could you please provide a reproducible sample code of the issue ?

henrymoulton:
Noticed Reanimated 3 has an unreleased PR related to a race condition  https://github.com/software-mansion/react-native-reanimated/pull/5224 - wonder if there's a relation to this issue.

beqramo:
Hi, thanks for your feedback, @gorhom.

I will try to provide more context before presenting the conclusions:


When snapToPosition is called from the ref right away after mount:
```tsx
  useEffect(() => {
        sheetRef.current?.snapToPosition(400);
  }, []);
```
At that time, SOMETIMES, the value of isLayoutCalculated is false. Why?

The value of isLayoutCalculated changes after the onLayout callback gets called (when the component is rendered and we have sizes and other details about the container). 
We are facing some kind of race condition; sometimes, onLayout is getting called first, and isLayoutCalculated and consecutively, animatedContainerHeight do have correct values.
It is like, in short, which one will get called first? useEffect or onLayout?

What are the solutions? (Sorted from worse to best solution)
1. To have some kind of callback prop like "onReady" that will indicate the parent component that the bottom-sheet is ready.
2. Return a specific value like "not_ready" when it will be triggered from the ref, right away after mount.
3. Take ownership of the problem from the library side; the library needs to automatically wait for the completion of all height calculation processes.


Reproducible sample code:
I first faced this problem on my app. When searching for solutions, I found that another developer already had the [same problem](https://github.com/gorhom/react-native-bottom-sheet/issues/1294).


Other than that I will drop my code snippet too:
```tsx
  const sheetRef = useRef<BottomSheet>(null);
  
  useEffect(() => {
        // call right after mount
        sheetRef.current?.snapToPosition(400);
  }, []);
  
return  <BottomSheet
      ref={sheetRef}
      snapPoints={[10]}
      index={ -1}
   >
      <Column mt={7} px={5} stretch>
             {/* children here*/}
      </Column>
    </BottomSheet>
```

Thank you. 

Hope I will not waste your time.

beqramo:
@henrymoulton Thanks for the info,
I think that one is a different issue from this.
From the funny side: I think I was facing that issue too and I'm happy that it is resolved and you informed me about it, waiting for the release.

Thank you

src/components/bottomSheet/BottomSheet.tsx
beqramo: I don't believe exiting from a function in that manner is correct. The caller has requested to open the bottom sheet, and the caller doesn't have any information on whether the package is ready or undergoing some heavy operations.

Regarding listening to isLayoutCalculated, if I understood it correctly, it seems more appropriate. I will update it after we know precisely whether there is a problem or not.

For more details on these matters, please refer to the comment section below.

henrymoulton:
@beqramo can you confirm it still happens if you're using Reanimated V2?

beqramo:
Hi @henrymoulton,

I am currently using my fork, and I haven't encountered the issue. Additionally, I am utilizing reanimated version 3. If you wish to test it, the problem might be more noticeable on low-end devices, particularly Android.

The issue you mentioned seems slightly different here. The problem lies in the synchronization of the `onLayout` and `snapToPosition` functions. The challenge arises because the first function is not synchronous, and its trigger can be delayed. That creates problem if `snapToPosition` gets called right away on parent component mount.

Thank you.

MBach:
Seems interesting! I'd like to see this PR merged if possible

jspizziri:
@gorhom , this is still an issue in `v5`. It's very difficult to reproduce as it depends on execution order so it can vary from device to device, platform to platform, and codebase to codebase.

It would seem that a simple solution could simply be to provide a callback on the BottomSheet like: `onReady` which triggers once `isLayoutCalculated` is `true`. Either that or queuing up calls to snap and then executing them once the `isLayoutCalculated` is `true`.

Issue #1689, opened 2024-01-09
[v4] | [v2] Unable to set initial index to be open


BethThomas141:
# Bug

We've just upgraded the bottom-sheet version in our app from v4.4.7 to v4.6.0. On doing this we now get the following error anywhere we use the bottom sheet: `'index' was provided but out of the provided snap points range! expected value to be between -1, 0`. We have four snap points passed into the drawer: -1 (hidden), 0 (closed), 1 (partially open), 2 (fully open).
We want the drawer to start partially open, so we want index = 1 on the Bottom sheet component. However, it seems like we can only pass in `index={-1}` or `index={0}` otherwise we see this error.

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.6.0   |
| react-native                    | 0.72.6   |
| react-native-reanimated         | 3.5.4   |
| react-native-gesture-handler    | 2.12.0   |

## Steps To Reproduce

1. Implement a bottom sheet and pass in array of at least two snapPoints. Pass in an index of 1.

Describe what you expected to happen:

1. The Bottom sheet starts at snapPoint 1 (height 600) on first load. Instead we get the error `'index' was provided but out of the provided snap points range! expected value to be between -1, 0`.

## Reproducible sample code

```
import React, { useMemo, useRef, useCallback } from 'react';
import { View, StyleSheet } from 'react-native';
import BottomSheet, { enableLogging } from '@gorhom/bottom-sheet';
import List from './components/List';
import SearchHandle from './components/SearchHandle';
import Button from './components/Button';

enableLogging();

const App = () => {
  // refs
  const bottomSheetRef = useRef(null);

  // variables
  const snapPoints = useMemo(() => [200, 400], []);

  // callbacks
  const handleExpandPress = useCallback(() => {
    bottomSheetRef.current?.expand();
  }, []);
  const handleCollapsePress = useCallback(() => {
    bottomSheetRef.current?.collapse();
  }, []);
  const handleClosePress = useCallback(() => {
    bottomSheetRef.current?.close();
  }, []);

  return (
    <View style={styles.container}>
      <Button onPress={handleExpandPress} title="Expand" />
      <Button onPress={handleCollapsePress} title="Collapse" />
      <Button onPress={handleClosePress} title="Close" />
      <BottomSheet
        ref={bottomSheetRef}
        snapPoints={snapPoints}
        keyboardBehavior="interactive"
[...]

PR #1694, opened 2024-01-10
fix: allow initial snapPoint index greater than 0

BethThomas141:
## Motivation

This is to fix bug #1689. 
Since version 4.5.0 there has been logic that if the container layout is not ready, the returned snapPoints are an array with just a single item `[-999]`. As a result, on initial load, before the container layout is ready, if you have passed in the index prop to the `BottomSheet` with a value greater than 0 (e.g. if you're passing in more than one snap point value, and have passed in  `index={1}` as you want the drawer to start open), it immediately errors as the index is out of range of the initial snapPoint array of `[-999]`. 

This PR reverts the initial load logic to how it was before version 4.5.0, where if the container layout is not yet ready, it returns instead an array filled with `-999` that matches the length of the passed in snapPoint array.


mlazari:
Any plans on merging this fix? I had to apply a patch with this fix since I was experiencing the same error.

PR #1700, opened 2024-01-19
Add missing mock of Touchables

jaworek:
## Motivation

Similarly to #1698 from @ghorbani-m. This PR adds missing mocks for Touchables. Without it when they're used in tested components, user will get this error:

`Warning: React.jsx: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.`


gorhom (approved):
thanks @jaworek for submitting this PR

PR #1727, opened 2024-02-03
fix documentation link

alexey-yarmosh:
fix documentation link


gorhom (approved):
thanks for fixing the link

PR #1741, opened 2024-02-15
refactor: add support for web without Babel/SWC

### LINKED ISSUE #1740: Enable support for Reanimated on the web without a Babel plugin
# Feature Request

Enabling [support on the web for Reanimated without using a Babel plugin](https://docs.swmansion.com/react-native-reanimated/docs/guides/web-support#web-without-the-babel-plugin) could help users who are using SWC (or other tooling) but would like to use this library.

I am going to need to `patch` this package regardless, so I would be happy to do a PR and get this into `5.0.0` (I'm on an alpha branch as it is).

## Why it is needed

> In order to use Reanimated without a Babel/SWC plugin, you need to explicitly pass the dependency array whenever you use a Reanimated hook.

## Possible implementation

We would need to add dependency arrays to anywhere in the codebase where they don't exist to these hooks:

- useDerivedValue
- useAnimatedStyle
- useAnimatedProps
- useAnimatedReaction

> Passing a dependency array is valid on both Web and native. Adding them will not negatively impact iOS or Android.

### Code sample

```
const sv = useSharedValue(0);
const dv = useDerivedValue(
  () => sv.value + 1,
  [sv] // dependency array here
);
```

joshsmith:
- refactor: add support for web without Babel/SWC
- refactor: add exhaustive deps rules to eslintrc for Reanimated hooks
- fix: add missing metro-react-native-babel-preset (breaking linting)

The changes simply involved looking through each use of the necessary Reanimated hooks mentioned in #1740 and replacing them with optional dependency arrays or fixing the dependency arrays to use the new exhaustive deps rule recommended by Reanimated in their documentation. You simply need to review that these changes were inclusive.

## Motivation

Closes #1740

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -195,7 +195,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
       return $modal
         ? _animatedContainerHeight.value - verticalInset
         : _animatedContainerHeight.value;
-    }, [$modal, topInset, bottomInset]);
+    }, [topInset, bottomInset, $modal, _animatedContainerHeight.value]);
     const animatedContainerOffset = useReactiveSharedValue(
       _providedContainerOffset ?? INITIAL_CONTAINER_OFFSET
     ) as Animated.SharedValue<Insets>;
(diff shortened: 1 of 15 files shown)


gorhom:
thanks @joshsmith for submitting this PR <3

src/components/bottomSheet/BottomSheet.tsx
```diff
@@ -274,7 +276,13 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
         isHandleHeightCalculated &&
         isSnapPointsNormalized
       );
-    });
+    }, [
+      _providedContainerHeight,
+      animatedContainerHeight.value,
+      animatedHandleHeight,
```
xilin, line 282: May I ask when shall we use `.value`?

PR #1750, opened 2024-02-22
Use function instead of component for BottomSheetModal children

Kerumen:
## Motivation

As `data` is an object, rendering `<Content data={data} />` creates a lot of unwanted re-render. Changing it to a function solves this issue.


Kerumen:
@gorhom Can you have a look? Thanks 🙏🏻

Kerumen:
Quick comment to remove the stale label..

gorhom (approved):

PR #1757, opened 2024-02-27
fix: fix bug on default animate config setting

jeongshin:
## Motivation

I'm testing bottom sheet v5 alpha.8 version. 
Then, I got error below. 

```
16:48:59 ERROR : ReanimatedError: Property 'ANIMATION_CONFIGS' doesn't exist, js engine: reanimated
```

Screenshot 2024-02-27 at 4 49 04 PM

I fixed to set default config under 'worklet', then it works fine. 

Thanks for great library! 🙌

### DIFF
--- a/src/utilities/animate.ts
+++ b/src/utilities/animate.ts
@@ -17,11 +17,15 @@ interface AnimateParams {
 
 export const animate = ({
   point,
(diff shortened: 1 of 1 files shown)


gorhom:
thanks @jeongshin for spotting and fixing the issue

Issue #1641, opened 2023-11-27
 [v4] When Talkback is enabled and a modal is displayed, the items below the modal sheet are still accessible by tap and the sheet is not focused


bekatd:
# Bug

1. When `Talkback` is enabled Modal is not focused when presented
2. Elements below the Modal are accessible
3. When dismissing modal, elements are no longer accessible/focusable (no green marker around them and no announcements)

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.5.1 |
| react-native                    | 0.72.3 |
| react-native-reanimated         | 3.5.4 |
| react-native-gesture-handler    | 2.13.4 |

## Steps To Reproduce

1. When `Talkback` is enabled Modal is not focused when presented
2. Elements below the Modal are accessible
3. When dismissing modal, elements are no longer accessible/focusable (no green marker around them and no announcements)

Describe what you expected to happen:

1. When `Talkback` is enabled, if user opens Modal, it should get focus automatically
2. Elements below the Modal should not be accessible by default (some prop needed maybe)
3. After dismissing Modal, elements that were below it should be accessible/focusable again

## Reproducible sample code
NA

instesti:
Yeah so this is still an issue, where the talkback skips children of the bottom-sheet and reads the backdrop instead. Any update on how to solve this? Currently it breaks a lot of accessibility wise

@gorhom

hraschan:
This is still an issue. It is possible to make the content reachable by setting the BottomSheetModal to `accessible={false}` as mentioned in #1764 . But the elements behind the Bottom Sheet are still accessible. Only appears on android.

I'm really looking forward for this getting fixed. Is it possible to reopen this issue? @gorhom

Hosam-hsm:
@hraschan @instesti  Were you able to find a solution to this issue?

hraschan:
Sadly not.

stevegreen99:
This issue occurs if dialogs are not launched using the showModal() method. It's best to implement the dialog using a <dialog> element because it has all the required behaviours built in. The role="dialog" attribute can also be used. If you do anything else, you will need to hide all the content outside the dialog using the "inert" attribute, but the dialog must not be inside an element with that attribute. This is all explained on the MDN website.

Tamimjabr:
Same problem here

Issue #687, opened 2021-10-08
[BottomSheetModal], a prop should allow elements below the sheet to not be accessible and the sheet should receive focus when presented


jsamr:
# Feature Request

I am implementing a feature for an app provided by the French state and accessibility is a top priority concern. The idea would be to reproduce the behavior of React Native `Modal` component whereby elements below the modal are not accessible.

The feature would be opt-in with a prop, `accessibiltiyHideSiblings` or something along those lines.

## Why it is needed

Screen reader users are confused when elements below the sheet are accessible. They are also confused when the focus is not passed to the modal, quoting Shopify accessibility team which shares conclusions of a study they made:

> When opening a modal window, focus must be placed on or inside of the modal window in order to bring context awareness to the user. Otherwise, focus remains on the activator control and the user may not be aware of or be able to easily reach the modal content.

Source: [Creating Accessible React Native Apps](https://www.shopify.com/partners/blog/react-native-accessibility)

## Possible implementation

The `<BottomSheetModalProvider>` which AFAIK acts like a portal could do the following:

- Hide siblings when modal is presented;
- Provide focus to the sheet when modal becomes presented.

### Code sample

``` jsx
function BottomSheetModalProvider() {
  // ...
  const isModalVisible = true;
  return (
    <>
      {/* Descendants */}
      <View
        importantForAccessibility={
          isModalVisible ? "no-hide-descendents" : "yes"
        }
      >
        {children}
      </View>

      <View
        importantForAccessibility={
          isModalVisible ? "yes" : "no-hide-descendents"
        }
        accessibilityViewIsModal={isModalVisible}
      >
        {/* Bottom sheet portal */}
      </View>
    </>
  );
}
```

This snippet does not include the second point (provide focus). This point can be handled by providing to the `View` which should receive focus the below function:

```jsx
/**
 * A function to pass as `onLayout` prop to focus on mount for screen readers.
 */
export function focusOnLayoutForScreenReaders(e: LayoutChangeEvent) {
  //@ts-expect-error - e.target is mistyped (it's actually a react element).
  const node = findNodeHandle(e.target);
  if (node) {
    AccessibilityInfo.setAccessibilityFocus(node);
    // Fix a bug on Android, see https://github.com/facebook/react-native/issues/30097
    AccessibilityInfo.setAccessibilityFocus(node);
  }
}
```

##### `accessibilityViewIsModal` (iOS)

[...]

PR #1288, opened 2023-02-08
Porting BottomSheet accessibility from v2 to v4

Mahmoud-SK:
## Motivation

I'm picking up this [task](https://github.com/gorhom/react-native-bottom-sheet/pull/693) of porting accessibility features from V2 to V4 as per [#548](https://github.com/gorhom/react-native-bottom-sheet/issues/548)

PR #1764, opened 2024-03-04
Fix: allow screen readers to detect modal content

flo-sch:
Hello @gorhom, thank you for this nice package :)

Here is a small contribution based on accessibility findings using TalkBack on Android and Voice Over on iOS.
If you are interested I could also port it to version 5

## Motivation

In the project I work on, accessibility support is a requirement. I noticed a few issues when integrating `<BottomSheetModal />` (we currently do not use `<BottomSheet />`):

1. Children cannot be read by Voice Over (iOS), only the modal itself is interpreted as a single "adjustable" view (my guess here is that elements with "accessibleRole: 'adjustable'" are not meant to have children? I could be wrong, this is just an assumption)
2. Text elements cannot be interpreted by Talk Back (Android) when using the default background, due to its accessibility props. This could easily be workaround by providing a custom background but I figured I could also propose this fix here.
3. I also face the issues described in #1641, but I did not manage to understand how to fix those myself unfortunately. I tried implementing some state in BottomSheetModalProvider (as initially proposed in #687), but since modals can be closed/dismissed in different ways, I could not implement a reliable state determining whether or not a modal is currently being presented (and adjust `importantForAccessibility` props based on that).

I recorded videos to illustrate what this PR fixes:

1. TalkBack

https://github.com/gorhom/react-native-bottom-sheet/assets/1577155/a7fa04c4-2e02-429a-a0f6-6e51d5644cd2
https://github.com/gorhom/react-native-bottom-sheet/assets/1577155/106c7c0c-c625-450a-ba29-3c34a3eb2927

2. Voice Over (unavailable in iOS simulator, but I used Accessibility Inspector instead)

https://github.com/gorhom/react-native-bottom-sheet/assets/1577155/dd0f8f89-71ec-49be-8d0a-09dd84943d6e
https://github.com/gorhom/react-native-bottom-sheet/assets/1577155/2d5ca920-d1e9-4afc-9b67-38122e016844

### DIFF
--- a/example/bare/android/build.gradle
+++ b/example/bare/android/build.gradle
@@ -3,10 +3,10 @@ import org.apache.tools.ant.taskdefs.condition.Os
 
 buildscript {
     ext {
-        buildToolsVersion = "31.0.0"
+        buildToolsVersion = "32.0.0"
         minSdkVersion = 21
-        compileSdkVersion = 31
-        targetSdkVersion = 31
+        compileSdkVersion = 32
+        targetSdkVersion = 32
 
         if (System.properties['os.arch'] == "aarch64") {
             // For M1 Users we need to use the NDK 24 which added support for aarch64
@@ -48,6 +48,11 @@ allprojects {
             }
         }
         google()
+        jcenter() {
+            content {
+                includeModule("com.eightbitlab", "blurview")
+            }
+        }
         maven { url 'https://www.jitpack.io' }
     }
 }
--- a/src/components/bottomSheetBackground/BottomSheetBackground.tsx
+++ b/src/components/bottomSheetBackground/BottomSheetBackground.tsx
@@ -7,13 +7,7 @@ const BottomSheetBackgroundComponent = ({
   pointerEvents,
   style,
 }: BottomSheetBackgroundProps) => (
-  <View
-    pointerEvents={pointerEvents}
-    accessible={true}
-    accessibilityRole="adjustable"
-    accessibilityLabel="Bottom Sheet"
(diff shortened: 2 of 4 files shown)


Eli-Nathan:
I think this can currently be overridden by passing `accessible={false}` to the BottomSheet.
Accessibility was only just recently added back in to this repo so you may need to update to at least `4.6.0`

https://github.com/gorhom/react-native-bottom-sheet/issues/1288
[Release notes](https://github.com/gorhom/react-native-bottom-sheet/releases/tag/v4.6.0)

flo-sch:
> I think this can currently be overridden by passing `accessible={false}` to the BottomSheet.
> Accessibility was only just recently added back in to this repo so you may need to update to at least `4.6.0`

Hi @Eli-Nathan, setting accessible to false fixes it for VoiceOver only (iOS). I included that in this PR (or well, setting it to `null` which works as well)

I already tried 4.6.0 and 4.6.1

flo-sch:
Un-stale

thienautran:
I've been able to get focus to child elements by using accessible=false, however still haven't been able to figure out how to make elements behind the bottom sheet inert, either with BottomSheetModal or regular BottomSheet. Anyone have any ideas on how to fix this?

flo-sch:
I believe it would require some implementation from this library first, as suggested [there](https://github.com/gorhom/react-native-bottom-sheet/issues/687#issue-1021133592). At least I have not found any way to achieve that yet

Hosam-hsm:
@flo-sch Did you find a solution?

PR #1784, opened 2024-04-01
refactor: refactored the snap points reactions and improve snapping sequencing

gorhom:
## Motivation

This PR introduces a fully rewritten snap points reactions handler which stabilised the snapping sequencing when multiple events trigger the snap mechanism.

The new handler insures that snap points changes does not block keyboard or any other triggers but to respect them and re-evaluate the bottom sheet position accordingly.

#### Use cases

This PR should fix

- on mount animation while parent container is resizing
- on mount animation with text input is auto focus 
- resizing snap points
- resizing content
- other more

##### Preview

https://github.com/gorhom/react-native-bottom-sheet/assets/4061838/398415f5-20d0-4f74-bdd6-cf61916e8c94

### DIFF
--- a/example/src/screens/advanced/KeyboardHandlingExample.tsx
+++ b/example/src/screens/advanced/KeyboardHandlingExample.tsx
@@ -73,6 +73,7 @@ const KeyboardHandlingExample = () => {
       <BottomSheet
         ref={bottomSheetRef}
         snapPoints={snapPoints}
+        enableDynamicSizing={false}
         keyboardBehavior={keyboardBehavior}
         keyboardBlurBehavior={keyboardBlurBehavior}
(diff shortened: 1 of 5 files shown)


x43n:
Looks great! Will it be possible to keep the keyboard open and input focused while updating the snap points with the drag handler? Like reddit/slack text input. Currently trying to do that with v5.alpha.9 but struggling to get it working.

mjvestal:
This PR fixes the issues I've found with adopting the `5.0.0-alpha.9` release in my React Native 0.72 app. Specifically with this PR applied I no longer see
* Sometimes a bottom sheet with only a focused text area is opens behind the keyboard
* Sometimes a bottom sheet opens to the wrong height

Great job!

Issue #1053, opened 2022-08-05
[v4] Bottom Sheet getting snapped to -1 (closed state) in certain situations with dynamic snap points


andrecrimb:
# Bug

I'm working in a project that has a bottom sheet with **2 dynamic snapPoints** (minimised, medium), when these are quickly changing, the bottom sheet snaps to -1 (closed state) and **doesn't snap back**. 
I observed that this issue mostly happens, if in one of these changes, the snapPoints aren't ascending.

### Video with sample code

https://user-images.githubusercontent.com/16760718/183125068-39815443-62ce-4203-b59d-4d7382c51815.mov

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.2.2   |
| react-native                    | 0.68.2   |
| react-native-reanimated         | 2.6.0   |
| react-native-gesture-handler    | 2.3.2   |

## Steps To Reproduce

- Bottom sheet needs to have 2 dynamic snapPoints
- SnapPoints need to be quickly changing.
- At some point snapPoints need to be in descending order 
 
**Use sample code provided bellow to reproduce this issue.**

Describe what you expected to happen:

1. `animatedIndex` snaps to -1 
2. Bottom sheet disappears and doesn't snap back 

## Reproducible sample code

```ts
import React, { useMemo, useRef } from 'react';
import { View, Text, StyleSheet } from 'react-native';
import BottomSheet from '@gorhom/bottom-sheet';
import { useSharedValue } from 'react-native-reanimated';

function generateRandomInt(min: number, max: number) {
  return Math.floor(Math.random() * (max - min) + min);
}

const useSimulateDynamicSnapPoints = () => {
  const [now, setNow] = React.useState(0);

  React.useEffect(() => {
    const interval = setInterval(() => {
      setNow(Date.now());
    }, 1);
    return () => {
      clearInterval(interval);
    };
  }, []);

  const snapPoints = useMemo(() => {
    const snapPointsNotAscending = [
      generateRandomInt(200, 250),
      generateRandomInt(100, 200),
    ];
    const snapPointsAscending = [
      generateRandomInt(100, 200),
      generateRandomInt(200, 250),
    ];

    if (Date.now() % 2 !== 0) {
      return snapPointsNotAscending;
    }
    return snapPointsAscending;
  }, [now]);
  return snapPoints;
};

const App = () => {
[...]

PR #1815, opened 2024-04-25
Fix BottomSheet closing instantly when snapPoints change

Maniae:

## Motivation

Probably related to https://github.com/gorhom/react-native-bottom-sheet/issues/1053

When snapPoints of a bottom sheet change while its animating, it may close it in some situations.
After some investigation, I think the reason for that is the animatedReaction on snapPoints/containerHeight that closes the bottom sheet under these conditions : `animatedCurrentIndex.value === -1` 

The problem being in some cases, when animation complete, the snapPoints change reaction seems to be called right after the completion callback, before any reaction on the animatedPosition. Since the the animatedPosition reaction is the one setting the animatedCurrentIndex value, this old value is used and causes the closing of the bottom sheet.

~~This fix is a simple line on the completion callback, setting the animatedCurrentIndex value to its goal value.~~
**The previous fix caused an issue where the onChange callback would not fire anymore.**

A better fix is to handle the case were the animation is STOPPPED in the snapPoints change listener, and set the new position using the actual animatedIndex value.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -1361,7 +1361,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
             animatedNextPositionIndex.value !== -1
               ? snapPoints[animatedNextPositionIndex.value]
               : animatedNextPosition.value;
-        } else if (animatedAnimationState.value === ANIMATION_STATE.STOPPED) {
+        } else if (animatedAnimationState.value === ANIMATION_STATE.STOPPED && animatedIndex.value % 1 === 0) {
           nextPosition = snapPoints[animatedIndex.value]
         } else if (animatedCurrentIndex.value === -1) {
           nextPosition = animatedClosedPosition.value;
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -665,6 +665,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
           },
         });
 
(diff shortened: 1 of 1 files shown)


XantreDev:
I think it's a good idea. Probably it will also fix issues with `enableDymanicSizing` resize

https://github.com/gorhom/react-native-bottom-sheet/assets/57757211/01a33a59-37ad-4c9a-8534-d6c696420bb4

ToledoSDL (approved):

theohdv:
Hey, do you know if it's going to be released soon in V4 ?

src/components/bottomSheet/BottomSheet.tsx
```diff
@@ -1361,6 +1361,8 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
             animatedNextPositionIndex.value !== -1
               ? snapPoints[animatedNextPositionIndex.value]
               : animatedNextPosition.value;
+        } else if (animatedAnimationState.value === ANIMATION_STATE.STOPPED && animatedIndex.value % 1 === 0) {
+          nextPosition = snapPoints[animatedIndex.value]
```
theohdv, line 1365:
```suggestion
          nextPosition = snapPoints[animatedIndex.value];
```

ryskin:
why closed? but is still appear

XantreDev:
Because of the genius bot

2ndMostGlossy:
after `snapToPosition` i set a timeout for 3 seconds before trying to set new snapPoints, this however caused the sheet to close. would that be a separate issue from this? My sense was that the animation was complete before I attempted to set the snap points so it would've been fine but that doesn't seem to be the case

Issue #1726, opened 2024-02-02
[v4] | BottomSheetFlatList not supported when newArchEnabled=true


vijaychouhan-rails:
# Bug

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.6.0   |
| react-native                    | 0.73.2   |
| react-native-reanimated         | 3.6.1  |
| react-native-gesture-handler    | 2.14.0  |

## Steps To Reproduce

1. Enable newArchEnabled=true
2. Add BottomSheetFlatList
3. Then open That bottom sheet and will thrown an error 
         

> Unexpected nativeTag: ${refType}; nativeTag=${nativeTag}
>          createBottomSheetScrollableComponent's ScrollableComponent needs to return 
> 	 a reference that contains a nativeTag to a Native HostComponent. ref=${ref}

And its throwing from this code [https://github.com/gorhom/react-native-bottom-sheet/blob/1128dc82e5c35e0846938e6007d0464b6bb35beb/src/utilities/getRefNativeTag.ts#L30](https://github.com/gorhom/react-native-bottom-sheet/blob/1128dc82e5c35e0846938e6007d0464b6bb35beb/src/utilities/getRefNativeTag.ts#L30)

Describe what you expected to happen:

1.  The error should not get and should work smoothly 

## Reproducible sample code

```
<BottomSheet
      ref={whatsappWindowRef}
      snapPoints={snapPoints}
      index={-1}
      enablePanDownToClose={true}
      backdropComponent={renderBackdrop}
      footerComponent={renderFooter}>
      <BottomSheetFlatList
        style={mainStyle.mB80}
        data={filteredData}
        keyExtractor={keyExtractor}
        renderItem={renderItem}
        ListHeaderComponent={title}
        ListEmptyComponent={emptyListComponent}
        ListFooterComponent={<View style={mainStyle.mB80} />}
      />
    </BottomSheet>
```

forever-sumit:
@gorhom can you please check this issue or any suggestion to solve that issue

vijaychouhan-rails:
@gorhom we are also waiting for that

vijaychouhan-rails:
@AndreiCalazans could you please check above issue

AndreiCalazans:
@vijaychouhan-rails since you have a reproducible at hand can you debug it and see how the library can retrieve the nativeTag when the new architecture is enabled?

[...]

PR #1823, opened 2024-05-02
fix: replace getRefNativeTag with findNodeHandle

AndreiCalazans:

## Motivation

Trying to fix this issue when new arch is enabled: https://github.com/gorhom/react-native-bottom-sheet/issues/1726

Current `getRefNativeTag` causes this issue with new architecture:

By reverting to findNodeHandle it works: 

https://github.com/gorhom/react-native-bottom-sheet/assets/20777666/ea29dda2-987a-47a1-a4c3-6a569a8916f6

We had introduced `getRefNativeTag` because experimental new architecture last year did not support using findNodeHandle. This is no longer true with the proxy renderer they created. 

This was tested with and without new architecture enabled on iOS.


gorhom:
tested v5 with/without new arch, it is not throwing this error

AndreiCalazans:
Really!? so it's unnecessary in v5, feel free to close the PR if so. @gorhom

techfoundrynz:
@gorhom can confirm v5 is not entirely working on new arch. I'm getting this after switching to new arch (also tried updating from alpha 9 to 10)
image

0xFA11:
it doesn't work with newArch=true even on 5.0.0-alpha.10 and RN 0.74.1 @gorhom 
image

brentvatne:
@ThusSpokeNomad @contactsimonwilson - please share a minimal reproducible example

0xFA11:
@brentvatne

`index.tsx`
```tsx
import { useState, useEffect } from "react";
import { Text } from "react-native";
import { GestureHandlerRootView } from "react-native-gesture-handler";
import WebView from "react-native-webview";
import BottomSheet, { BottomSheetScrollView } from "@gorhom/bottom-sheet";

export default () => {
    const [lipsum, setLipsum] = useState<string>();
    useEffect(() => {
        void (async () => {
            const lipsumJSON = await (await fetch("https://www.lipsum.com/feed/json")).json();
            setLipsum(lipsumJSON.feed.lipsum);
        })();
    }, []);
    return (
        <GestureHandlerRootView>
            <WebView source={{ uri: "https://www.lipsum.com/feed" }} />
            <BottomSheet snapPoints={["50%", "100%"]}>
                <BottomSheetScrollView>
                    <Text>{lipsum}</Text>
                </BottomSheetScrollView>
            </BottomSheet>
        </GestureHandlerRootView>
    );
};
```

`package.json`
```json
{
    "name": "minrepro",
    "main": "expo-router/entry",
    "scripts": {
        "android": "expo run:android",
        "ios": "expo run:ios"
    },
    "dependencies": {
        "@gorhom/bottom-sheet": "^5.0.0-alpha.10",
        "expo": "^51.0.9",
        "expo-build-properties": "0.12.1",
        "expo-router": "3.5.15",
        "react": "18.2.0",
        "react-native": "0.74.1",
        "react-native-gesture-handler": "2.16.1",
        "react-native-reanimated": "3.10.1",
        "react-native-safe-area-context": "4.10.1",
        "react-native-webview": "13.8.6"
    },
    "devDependencies": {
        "@types/react": "18.2.79",
        "typescript": "5.3.3"
    }
}
```

also, interestingly, if you delete `useState` hook then it works fine (and `useEffect` could just `console.log` for testing).

janicduplessis:
I've also hit this issue on latest v5 and RN 0.74, it seems to happen only when using `BottomSheetScrollView`. This patch fixes it for me.

janicduplessis (commented):
Probably also want to delete the getRefNativeTag.web.ts file.

brentvatne:
cc @gorhom re: @janicduplessis' comment - should we proceed with this pr?

AndreiCalazans:
@janicduplessis I don't see a getRefNativeTag.web in the source code.

janicduplessis:
Looks like it is only in the v5 branch, nvm then!

craftzdog:
confirmed it solved the issue in my project with RN0.74.2 + New Arch

brentvatne:
cc @gorhom - wdyt?

gorhom:
@AndreiCalazans sorry i have been very busy with work lately,, just to confirm is this fix backward compatible with old arch ?

AndreiCalazans:
@gorhom Yes. This proxy we included was a workaround for the new arch when it didn't have support for findNodeHandle - no longer required.

gorhom:
thanks for confirming @AndreiCalazans , this fix should be out for the next releases of `v4` and `v5`

PR #1864, opened 2024-06-17
New public method getState. Easier to open and close sheet with one button

BogdanGorelkin:
In some cases it would be useful to get the BottomSheet state. Especially when the same button should open and close the sheet. It would be great if you'll add this feature to the new version.

P.S: The same changes made from the [HEAD v5.0.0-alpha.10](https://github.com/gorhom/react-native-bottom-sheet/tree/v5.0.0-alpha.10) can be found in [this brunch](https://github.com/BogdanGorelkin/react-native-bottom-sheet/tree/public_method_getState-v5.0.0-alpha.10).

PP.S: I temporarily commented out the ignoring of the _/lib_ folder in the _.gitignore_ file. Without this change while installing package by command: `npm install BogdanGorelkin/react-native-bottom-sheet#public_method_getState-v5.0.0-alpha.10`  types are not being compiled


BogdanGorelkin:
This pull request is still actual

gorhom:
Hi @BogdanGorelkin,

1. this PR has 516 files change, which i won't accept, i hope you understand the rational here.
2. providing animated index to the bottom sheet will always update you with the bottom sheet position index, which will tells you if the bottom sheet is open or close

PR #1897, opened 2024-07-26
feat: add stack behavior replace-and-dismiss

janodetzel:
I added a new stack behavior type which dismisses the previous modal instead of minimizing it.

## Motivation

This new stack behaviour type is important to enable users to dismiss the previous modal when a different modal is opened.

### DIFF
--- a/src/components/bottomSheetModalProvider/BottomSheetModalProvider.tsx
+++ b/src/components/bottomSheetModalProvider/BottomSheetModalProvider.tsx
@@ -45,18 +45,17 @@ const BottomSheetModalProviderWrapper = ({
       }
 
       /**
-       * Minimize the current sheet if:
+       * Dismiss or minimize the current sheet if:
        * - it exists.
        * - it is not unmounting.
        * - stack behavior is 'replace'.
        */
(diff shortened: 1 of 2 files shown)


janodetzel:
@gorhom can you have a look at this new stack behaviour type? Would triggering `dismiss()` instead of `minimize()` cause any problems?

gorhom:
hi @janodetzel thanks for submitting this PR, i think the naming `replace` does not represent its functionality , might worth renaming it to `stack` or `switch`. and `replace` would be the present and dismiss old sheet

EvgenMart:
@gorhom , Greetings! This may not be entirely on topic, but I would like to ask. Is there a possibility to add a current example of using BottomSheet with @react-navigation/native? The link to the example in the description is no longer valid: https://github.com/gorhom/react-native-bottom-sheet/blob/master/example/bare/src/screens/integrations/NavigatorExample.tsx

gorhom:
@janodetzel thanks for submitting this PR, i have already introduce this feature on v5 https://gorhom.dev/react-native-bottom-sheet/modal/props#stackbehavior

PR #1951, opened 2024-09-27
Docs: Update URLs in README

mihnea-popescu:
The URLs inside the README.md were not updated.

## Motivation

I have updated the URLs leading to the documentation in the README.md to the current ones.

For example, https://ui.gorhom.dev/components/bottom-sheet was not redirecting correctly.


gorhom:
thanks for updating the links @mihnea-popescu

PR #1969, opened 2024-10-14
fix: auto-close action

dannyhw:
Not sure how to test this but I think the check in auto close is always returning false and so always closing issues. Probably because its `true` not `'true'`

saw a similar example here
https://github.com/orgs/community/discussions/26712#discussioncomment-3253013

## Motivation

All issues being created right now are automatically closed even when following the template correctly


dannyhw:
already fixed

gorhom:
thanks @dannyhw, i went ahead and fixed the issue and i added extra checks if users inserted link to the lib docs

PR #1971, opened 2024-10-14
fix: add children to BottomSheetModal containerComponent type

Nodonisko:
## Motivation

`children` as default prop of `React.FC` was removed in React 18 so it needs to be explicitly defined. Otherwise this code throws error:

```tsx
        <BottomSheetModal
             {...otherProps}
            // this will throw TS error
            containerComponent={({ children }) => <WindowOverlay>{children}</WindowOverlay>}
        >
```

### DIFF
--- a/src/components/bottomSheetModal/types.d.ts
+++ b/src/components/bottomSheetModal/types.d.ts
@@ -43,7 +43,7 @@ export interface BottomSheetModalProps
    * @type React.ComponentType
    * @default undefined
    */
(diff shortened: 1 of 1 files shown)


gorhom:
thanks @Nodonisko for submitting this PR, i think it would be better to use
```
  containerComponent?: React.ComponentType<React.PropsWithChildren>;
```

wdyt ?

Nodonisko:
Sounds better, fixed.

PR #1848, opened 2024-05-28
fix: assigning to read-only property 'reduceMotion'

pafry7:
## Motivation

Upgrading the package to version 4.6.3 results in the following error when rendering BottomSheet in the development.

```
ERROR  TypeError: Cannot assign to read-only property 'reduceMotion'

This error is located at:
    in AnimatedComponent(View)
    in Unknown (created by PanGestureHandler)
    in PanGestureHandler (created by BottomSheetDraggableViewComponent)
    in BottomSheetDraggableViewComponent (created by BottomSheet)
    in RCTView (created by View)
    in View (created by AnimatedComponent(View))
    in AnimatedComponent(View)
    in Unknown (created by BottomSheet)
    in RCTView (created by View)
    in View (created by AnimatedComponent(View))
    in AnimatedComponent(View)
    in Unknown (created by BottomSheet)
    in RCTView (created by View)
    in View (created by BottomSheetContainerComponent)
    in BottomSheetContainerComponent (created by BottomSheet)
    in BottomSheetGestureHandlersProvider (created by BottomSheet)
    in BottomSheet (created by BottomSheet)
    ...
    in AppContainer
    in main(RootComponent), js engine: hermes
```

## Cause

Function `animate` inside `src/utilities/animate.ts` uses `worklet` and you cannot modify the objects inside the `worklet`, as explained here: https://github.com/software-mansion/react-native-reanimated/issues/5430#issuecomment-1831453991.

## Solution
We have to add `reduceMotion` if:
- ReduceMotion exists (it does not in earlier versions of react-native-reanimated)
- the passed config is not undefined (to preserve current behaviour with default configs)

It would be better to do it in one place or we can extract the logic to the utility function

## env-info
```
  expo-env-info 1.2.0 environment info:
    System:
      OS: macOS 14.4.1
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 20.11.1 - ~/.volta/tools/image/node/20.11.1/bin/node
      Yarn: 1.22.22 - ~/.volta/tools/image/yarn/1.22.22/bin/yarn
      npm: 10.7.0 - ~/projects/alergeek/intu/app/node_modules/.bin/npm
      Watchman: 2023.12.04.00 - /opt/homebrew/bin/watchman
    Managers:
      CocoaPods: 1.14.3 - /Users/frydson/.rbenv/shims/pod
    SDKs:
      iOS SDK:
[...]

PR #1975, opened 2024-10-15
fix: cannot add new property 'reduceMotion'

fobos531:
Hello!

In https://github.com/gorhom/react-native-bottom-sheet/pull/1743 I introduced a fix for the sheet not appearing because of reduce motion settings. However, my implementation was faulty and caused a very annoying issue, which was fixed by @pafry7 flawlessly in https://github.com/gorhom/react-native-bottom-sheet/pull/1848. After I updated to v5, this issue started reappearing. When I looked at the source code for v5, I noticed that you preserved my implementation. This PR changes the implementation back to the one from @pafry7 one, so all credits go to them. 

I wasn't able to reproduce the issue in your example app, but I have confirmed locally that this diff works in my production app in which I was able to consistently reproduce it (it usually happened after I opened a sheet with dynamic sizing enabled, and then opened another one which had dynamic sizing disabled). I hope you can review and merge this PR.

Thank you!

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -22,6 +22,7 @@ import Animated, {
   useWorkletCallback,
   type WithSpringConfig,
   type WithTimingConfig,
+  ReduceMotion,
 } from 'react-native-reanimated';
 // import BottomSheetDebugView from '../bottomSheetDebugView';
 import {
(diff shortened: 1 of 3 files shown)


gorhom:
thanks @fobos531 for submitting this PR, but i decided to introduce a different approach, where i do not want to override this property by default, but allow users to override it when needed.

https://github.com/gorhom/react-native-bottom-sheet/commit/1ef05c7fee821c356220452ccf61d33d29483c00

read more: https://gorhom.dev/react-native-bottom-sheet/props#overridereducemotion

PR #1988, opened 2024-10-21
fix: added BottomSheetFlashList mock

Fadikk367:
## Motivation

Current `mock.js` file lacks export for `BottomSheetFlashList`. Please let me know if simply assigning `ReactNative.FlatList` to it is fine or if I should try to conditionally import a real mock from `@shopify/flash-list`.


gorhom (approved):
thanks @Fadikk367

Issue #1983, opened 2024-10-18
Address reanimated API misuse warnings that pop up starting with reanimated 3.16.0


vestrelatlassian:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS, Android

### What happened?

Reanimated version [3.16.0](https://github.com/software-mansion/react-native-reanimated/releases/tag/3.16.0) introduced a logger that by default shows warnings about API misuse.

![Screenshot 2024-10-17 at 3 49 45 PM](https://github.com/user-attachments/assets/59f0f765-35cd-4530-aa97-e54b9183caa6)

We are currently using bottom sheet version `5.0.2` in our app and noticed warnings started to pop up whenever the bottom sheet is rendered/re-rendered.

![Screenshot 2024-10-17 at 3 38 01 PM](https://github.com/user-attachments/assets/9d377915-4965-4708-9ff2-96c9d336cd6c)

### Reproduction steps

1. Upgrade reanimated lib to version [3.16.0](https://github.com/software-mansion/react-native-reanimated/releases/tag/3.16.0) or newer.
2. Render bottom sheet.

### Reproduction sample

https://snack.expo.dev/IUep5KqJQjKmj-dPQdc3g

### Relevant log output

```shell
If you don't want to see this message, you can disable the `strict` mode. Refer to:
https://docs.swmansion.com/react-native-reanimated/docs/debugging/logger-configuration for more details.

(NOBRIDGE) WARN  [Reanimated] Reading from `value` during component render. Please ensure that you do not access the `value` property or use `get` method of a shared value while React is rendering a component.
```

nicolaslevavasseursmile:
Same here

ghost:
I am also facing this issue 

![Screenshot 2024-10-18 at 4 07 31 PM](https://github.com/user-attachments/assets/f0a32d0f-ffcd-44d8-bb09-5a57f48e2ccb)

FernandoAOborges:
same for me. 
<img src="https://github.com/user-attachments/assets/dbc12d53-18b6-4ccf-81b3-36dec2e92487" width="200" />

To resolve this for now, create a file named **reanimatedConfig.js** at the root of your project, then import it in App.tsx.

```
import {
  configureReanimatedLogger,
  ReanimatedLogLevel,
} from 'react-native-reanimated';

configureReanimatedLogger({
  level: ReanimatedLogLevel.warn, 
[...]

PR #1992, opened 2024-10-23
fix(#1983): updated shared values access as hook dependancies

pinpong:

## Motivation


gorhom:
thank you @pinpong

PR #2006, opened 2024-10-29
fix: stop any running animation, before early exiting due to same position

ghitier:
Fixes an edge case of `animateToPosition`, where it would early exit when `position === animatedPosition.value`, but wouldn't stop the currently running animation.
Therefore the bottom-sheet would proceed animating to an erroneous position.


73k05:
Up please I also need this fix for my app :-)

ghitier:
This PR still fixes an edge case and shouldn't be marked stale.

PR #2008, opened 2024-10-29
Update mock.js for new reanimated functions

JorenVerbaandert:
Update the mock to work with react-native-reanimated 3.16

## Motivation

Allow the mock to work correctly when using the new `get` and `set` api in reanimated


gorhom:
thanks @JorenVerbaandert

PR #2058, opened 2024-11-21
fix[doc]: Change references to react-native-gesture-handler from v3 to v2

pyramid-scheme-ceo:
Change references to react-native-gesture-handler from v3 to v2

## Motivation

Went to upgrade to v5 and noticed the mentions of react-native-gesture-handler v3 on the homepage so naturally I went to upgrade that first ...only to realise RNGH is still only on v2 (which is also mentioned in the blog post here: https://gorhom.dev/react-native-bottom-sheet/blog/bottom-sheet-v5). Correct me if I'm wrong but I think the docs are supposed to say v2 where they currently say v3.

### DIFF
--- a/website/docs/index.md
+++ b/website/docs/index.md
@@ -64,7 +64,7 @@ npx expo install react-native-reanimated react-native-gesture-handler
 ```
 
 :::info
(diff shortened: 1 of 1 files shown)


qnrjs42:
react-native-gesture-handler v3 didn't come out, and I was confused for a while about how to install it

gorhom:
thanks for submitting this PR

PR #2059, opened 2024-11-25
feat: A11y Props Support

zaferatli:
I added a11y props support and removed magic a11y props.

## Motivation

You may know the 3rd December is the World Disability Day and I'm enhancing to a11y ux to my own app and magic a11y create issue to us and many other props not able to settled.


brandondalbec:
+1 on this, would be extremely helpful as we're also trying to make our app more accessible.

Also worth noting - For users using physical keyboards/external assistive technology, the backdrop on iOS is completely inaccessible and prevents the user from getting out of it. They become completely stuck.

zaferatli:
up up up @gorhom

zaferatli:
aq yöneteceğiniz reponun

zaferatli:
aq yöneteceğiniz reponun

PR #2072, opened 2024-12-06
fix: prevent multiple calls for onClose event

wawrow:
currently `_providedOnClose` gets called potentially multiple times as this reanimated hook may (and likely will) run multiple times - I believe it should be called only once when changing `_animatedIndex` to `-1` only

## Motivation

if you perform some action `onClose`it may trigger multiple times.


wawrow:
> This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

I'd rather have that merged; @gorhom can you please help with this?

Issue #2043, opened 2024-11-17
[Bug]: overwriting BottomSheet.onAnimate will reset snap index to 0 on rerender


Jacky56:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

Android, Web

### What happened?

overwriting `BottomSheet.onAnimate` will reset snap index to 0 on rerender

### Reproduction steps

overwrite `onAnimate` method for `BottomSheet` with whatever then create an event which causes a rerender

### Reproduction sample

https://snack.expo.dev/@kagutron/bottom-sheet---issue-reproduction-template

hanwenbo:
Yes, there has been bug, you need to change the source code, when you operate fast, it will appear, you can prompt the user on the interface, please slow down  ：D

IslamRustamov:
The issue lies in `handleOnAnimate` function being a dependency for `animateToPosition` in @gorhom/bottom-sheet/src/components/bottomSheet/BottomSheet.tsx

I didn't manage to find the reason how a dependency change retriggers `animateToPosition` (which retriggers something else and so on). Will look into that if I have time. But please do join me in this bugfixing.

UPDATE:
I found the reason why that is happening. **handleOnAnimate** recreates, which recreates **animateToPosition**, which recreates **handleSnapToIndex**, which is calling **useEffect** on this [line](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L1862) . Which is basically why this bug happens. Will think of a fix for that.

UPDATE 1:
Here is the fix https://github.com/gorhom/react-native-bottom-sheet/pull/2073

IslamRustamov:
bump

gorhom:
thanks @IslamRustamov for your investigation and proposed fix. This issue should be resolved with the next release

SoheilKhorram:
> thanks [@IslamRustamov](https://github.com/IslamRustamov) for your investigation and proposed fix. This issue should be resolved with the next release

@gorhom on what date will this update be done? when will be the next release

gorhom:
it should be published now https://github.com/gorhom/react-native-bottom-sheet/releases/tag/v5.1.1

PR #2073, opened 2024-12-07
fix: handle unnecessary invocation of handleSnapToIndex on onAnimate …

IslamRustamov:
…recreation

I create a ref which stores previous passed initial index and if [useEffect](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L1862) is invoked again it checks whether the new **_providedIndex** actually does not equal to the old **_providedIndex**

## Motivation

https://github.com/gorhom/react-native-bottom-sheet/issues/2043
When we pass a prop **onAnimate** to **BottomSheet** we will encounter a bug, where **onAnimate**'s function recreation will cause a chain of dependency changes (**onAnimate** -> **_providedOnAnimate** -> **handleOnAnimate** -> **animateToPosition** -> **handleSnapToIndex** -> [useEffect](https://github.com/gorhom/react-native-bottom-sheet/blob/master/src/components/bottomSheet/BottomSheet.tsx#L1862)) which will cause our **BottomSheet** to snap back to the initial index (which is a wrong behaviour).

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -6,6 +6,7 @@ import React, {
   useImperativeHandle,
   memo,
   useEffect,
+  useRef,
 } from 'react';
 import { type Insets, Platform } from 'react-native';
 import { State } from 'react-native-gesture-handler';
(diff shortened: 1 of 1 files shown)


IslamRustamov:
bump

erickreutz:
LGTM!

gorhom:
thanks @IslamRustamov, for submitting this PR, it indeed fixes the issue, but it did not address the root cause of the issue, which is, when providing an unstable callback `onAnimate`, the internal method `handleSnapToIndex` will get re-created every time.

i have created a commit inspired by your solution and investigation, to make `handleSnapToIndex` as an event callback rather than a callback, by using `useStableCallback`

PR #2114, opened 2025-01-15
docs: fix mentioned gesture handle version

gustavoggs:

## Motivation

I didn't found a v3 of React Native Gesture Handler library, so I believe we want to mention v2 instead of v3 in this part of the docs.


gorhom:
thanks for submitting the PR , it is already fixed by https://github.com/gorhom/react-native-bottom-sheet/pull/2058

PR #2133, opened 2025-01-27
fixed bottom-sheet getting closed in initial render

AyushMittal42:
## Motivation

So we faced an issue in which BottomSheet gets closed immediately after appearing.
I debugged it and found the root cause which is a race condition i.e. our initial render animation got finished and at the same time Snap-points changes
Now what happens is before the `animatedCurrentIndex` got changes to 1, evaluatePosition function got triggered due to snap changes and inside it goes in the closed condition block (since `animatedCurrentIndex` is still -1) and this causes the BottomSheet to get in closed state

### Fix
So for fixing this I simply added another check which is for `animatedNextPositionIndex` to check whether it is also -1 which will make sure we want that BottomSheet to be in closed state and then only we will trigger this closing

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -976,7 +976,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
          */
         if (
           animatedAnimationState.value !== ANIMATION_STATE.RUNNING &&
-          animatedCurrentIndex.value === -1
+          animatedCurrentIndex.value === -1 && animatedNextPositionIndex.value === -1
         ) {
           /**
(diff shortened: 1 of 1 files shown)


ryskin:
I applied this fix, but the BottomSheet doesn’t lock — it stops without fully expanding and doesn’t close either. I’ve been trying to fix this bug for months. 😩

ryskin:
I applied this fix, the BottomSheet no longer collapses, but it stops without fully expanding. I’ve been trying to fix this bug for months. 😩

AyushMittal42:
@ryskin This change only prevent closing of the BottomSheet. How does it affect the opening?
Did you made the exact same changes or added something else as well?

AyushMittal42:
@gorhom can you please take a look on this?

PR #2160, opened 2025-02-13
feat: add enabled prop to draggable view

geovie:
## Motivation

The PR adds a simple `enabled` flag to `BottomSheetDraggableView` which sets the enabled flag on the gesture.   
Previously `enableContentPanningGesture` was accessed via `useBottomSheetInternal`.

By allowing to set the `enabled` flag on the component directly it is now possible to use `BottomSheetDraggableView` even if the `enableContentPanningGesture` of the `BottomSheet` is set to `false`  and so `BottomSheetDraggableView` can be used to only make part of the sheet's content draggable.

 Another side effect is the removal of:
 ```typescript
 const DraggableView = enableContentPanningGesture
      ? BottomSheetDraggableView
      : Animated.View;
 ```
 which previously caused a remount of the content when `enableContentPanningGesture` was changed.

### DIFF
--- a/example/src/screens/advanced/DraggableViewExample.tsx
+++ b/example/src/screens/advanced/DraggableViewExample.tsx
@@ -1,0 +1,74 @@
+import React, { useCallback, useMemo, useRef } from 'react';
+import { View, StyleSheet, Text } from 'react-native';
+import BottomSheet, { BottomSheetDraggableView, BottomSheetView } from '@gorhom/bottom-sheet';
+import { Button } from '../../components/button';
+
+const DraggableViewExample = () => {
+  // hooks
+  const bottomSheetRef = useRef<BottomSheet>(null);
+
+  // variables
+  const snapPoints = useMemo(() => ['25%', '50%'], []);
+
+  // callbacks
+  const handleExpandPress = useCallback(() => {
+    bottomSheetRef.current?.expand();
+  }, []);
+  const handleCollapsePress = useCallback(() => {
+    bottomSheetRef.current?.collapse();
+  }, []);
+  const handleClosePress = useCallback(() => {
+    bottomSheetRef.current?.close();
+  }, []);
+
+  // renders
+  return (
+    <View style={styles.container}>
+      <Button label="Expand" onPress={handleExpandPress} />
+      <Button label="Collapse" onPress={handleCollapsePress} />
+      <Button label="Close" onPress={handleClosePress} />
+      <BottomSheet
+        ref={bottomSheetRef}
+        snapPoints={snapPoints}
+        keyboardBehavior="interactive"
+        keyboardBlurBehavior="restore"
+        enablePanDownToClose={true}
+        enableContentPanningGesture={false}
+      >
+        <BottomSheetView style={styles.row}>
+          <BottomSheetDraggableView style={styles.leftView}>
+            <Text>Draggable</Text>
+          </BottomSheetDraggableView>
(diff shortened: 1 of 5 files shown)


src/components/bottomSheet/BottomSheet.tsx
```diff
@@ -1906,12 +1903,13 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
                   accessibilityRole={_providedAccessibilityRole ?? undefined}
                   accessibilityLabel={_providedAccessibilityLabel ?? undefined}
                 >
-                  <DraggableView
+                  <BottomSheetDraggableView
                     key="BottomSheetRootDraggableView"
                     style={contentContainerStyle}
+                    enabled={enableContentPanningGesture}
```
gorhom, line 1909: i dont think we need to pass it as props, as it is already read internally from `useBottomSheetInternal`

gorhom (changes requested):
thanks for submitting this PR, please address the comment i left

src/components/bottomSheet/BottomSheet.tsx
geovie: This is necessary so that `BottomSheetDraggableView` can also be used when `enableContentPanningGesture` of the `BottomSheet` is set to `false`, see PR description.

I've added an example `DraggableViewExample` which showcases the usage.

gorhom: mmm i see, i think for scalability , i would move the `enableContentPanningGesture` to each scrollable view instead. but we could go ahead with this temporary solution

geovie: Yeah, in my case I need to wrap an external component so the solution of this MR would work better for me...

IslamRustamov:
@gorhom when do you plan to release this change? Since I'm facing an issue that this exact PR fixes and need to know whether I should patch it locally or wait for a new release.

example/src/screens/advanced/DraggableViewExample.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <BottomSheet
+        ref={bottomSheetRef}
+        snapPoints={snapPoints}
+        keyboardBehavior="interactive"
+        keyboardBlurBehavior="restore"
+        enablePanDownToClose={true}
+        enableContentPanningGesture={false}
+      >
+        <BottomSheetView style={styles.row}>
+          <BottomSheetDraggableView style={styles.leftView}>
```
IslamRustamov, line 39: I think you miss `enabled` prop here, no?

Edit: it's set as default as true, okie

bestofaIl:
@gorhom, do you plan to merge it, or is there a better solution that will be presented in future releases?

Uziniii:
any update on the merge status ?

gomezger:
any news about that @gorhom @geovie @IslamRustamov ?

ucheNkadiCode:
Did anyone create a patch package for this by chance? I'm not understanding how to merge these changes with the latest version of this package

src/components/bottomSheet/BottomSheet.tsx
ucheNkadiCode: Is this something that we should / could create a patch package around? I'm not sure if this is still being worked on but could solve some issues I'm having while trying to implement a draggable flatlist inside of this <bottomsheet> component

PR #2162, opened 2025-02-14
Update docs to use BottomSheetView

jeremymelchor:
I've updated the doc to use the `BottomSheetView` instead of simple `View`


jeremymelchor:
up

PR #2191, opened 2025-03-07
feat(a11y): transfer accessible property to BottomSheet's children

thomas-cicognani-post:
Transfer accessible property to Bottom Sheet's children in order to improve keyboard navigation and VoiceOver / TalkBack results.

## Motivation

When fixing some accessibility features like keyboard navigation in our application, I discovered that some parts of the Bottom Sheet are accessible when you use, for instance, the keyboard tabulation. The consequence is users can be lost because the selected element in UI is not visible nor pertinent. Same for blind or visual impaired people who use VoiceOver / TalkBack, the smartphone says thinks like "Bottom Sheet" or "Bottom Sheet Handle", which may be not pertinent to browse the app.

In that case, the main property "accessible" from the Bottom Sheet props may be transferred to its children if we need to enable or disable the accessibility. We only changed components we use and must be updated with the accessible property to fix our app.

Thus, in our app, we instantiate several Bottom Sheets, so we have to disable all hidden sheets and only enable the one which is opened. If we don't do that, keyboard tabulation goes through all Bottom Sheets even they are not visible.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -1883,6 +1883,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
             />
             <BottomSheetContainer
               key="BottomSheetContainer"
+              accessible={_providedAccessible ?? undefined}
               shouldCalculateHeight={!$modal}
               containerHeight={_animatedContainerHeight}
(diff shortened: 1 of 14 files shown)


thomas-cicognani-post:
Hello. Is it possible to not close this PR?
I think it will be very appreciable to improve keyboard navigation for persons with disabilities.

PR #2219, opened 2025-03-28
chore: add forwardRef to all eligible components

bryanmylee:
## Motivation

When using styling libraries like Tamagui, it's often useful to wrap components like `<BottomSheetView>` with the `styled` function. However, these component wrapper functions require the use of `forwardRef` on their targets.

When viewing the source, there does not appear to be any reason why `forwardRef` cannot be used for many components. This PR simply wraps all eligible components and passes the `forwardedRef` to the inner components.

### DIFF
--- a/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
+++ b/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
@@ -1,4 +1,5 @@
 import React, {
+  forwardRef,
   memo,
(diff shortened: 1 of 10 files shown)


bryanmylee:
Small bump on this, I've been using a patch for this and there have been no issues so far.

bryanmylee:
Can we reopen this? Unless we're just accepting that all users will be on React 19 in which case `forwardRef` is no longer needed.

PR #2223, opened 2025-04-01
Update peer dependency for reanimated to allow the use of reanimated v4

skusnierz:
## Motivation

Currently, running an application that uses react-native-bottom-sheet along with reanimated v4 causes a warning to be displayed :
```
The following packages should be updated for best compatibility with the installed expo version:
  react-native-reanimated@4.0.0-beta.3 - expected version: ~3.16.1
Your project may not work correctly until you install the expected versions of the packages.
```
The following changes will cause this warning not to be displayed, I tested with the example that is in the repository and everything works fine

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "react": "*",
     "react-native": "*",
     "react-native-gesture-handler": ">=2.16.1",
-    "react-native-reanimated": ">=3.16.0"
+    "react-native-reanimated": ">=3.16.0 || >=4.0.0-"
   },
   "peerDependenciesMeta": {
     "@types/react-native": {


package.json
```diff
@@ -65,7 +65,7 @@
     "react": "*",
     "react-native": "*",
     "react-native-gesture-handler": ">=2.16.1",
-    "react-native-reanimated": ">=3.16.0"
+    "react-native-reanimated": ">=3.16.0 || >=4.0.0-"
```
skusnierz, line 68: fyi: this `-` at the end is not typo, it is about beta version coverage e.g. `4.0.0-beta.3`

tomekzaw:
@gorhom Can we please re-open this PR?

tomekzaw:
We will also need to replace all `useWorkletCallback` calls with `useCallback(() => { 'worklet'; })` because `useWorkletCallback` has been removed in Reanimated v4.

edit: Done in https://github.com/gorhom/react-native-bottom-sheet/pull/2356

gorhom:
thanks @skusnierz for submitting this PR

lucaswitch:
> We will also need to replace all `useWorkletCallback` calls with `useCallback(() => { 'worklet'; })` because `useWorkletCallback` has been removed in Reanimated v4.
> 

https://github.com/user-attachments/assets/5e713c63-8aef-494b-be04-b74c53949201


> edit: Done in #2356

Also the example app is not quite working with reanimated v4. I'm not sure if this should be reported to reanimated or this repo cause it works fine on v3. It seems like a behavior change not reported on migration guide between v3 and v4.

PR #2244, opened 2025-04-17
fix: prevent canceling touchmove events when not cancelable

erickreutz:
## Description

This PR fixes warnings about attempting to cancel touchmove events with cancelable=false by adding checks for the event.cancelable property before calling preventDefault().

## Details

The error 'Ignored attempt to cancel a touchmove event with cancelable=false, for example because scrolling is in progress and cannot be interrupted' occurs when the code tries to call preventDefault() on a touchmove event that has cancelable=false (which happens during active scrolling).

The fix adds event.cancelable checks in all three places where preventDefault() is called in the useScrollHandler.web.ts file.


erickreutz:
not state

gorhom:
thanks @erickreutz for submitting this PR

Issue #2237, opened 2025-04-09
[Bug]: findNodeHandle i not supported on web and this is crashing all react native web apps


idrisssakhi:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

Web

### What happened?

The bottom sheets do not work anymore on web on latest expo + react native versions

### Reproduction steps

* have any bottom sheet component rendered on react-native web app
* expo 52.0.44

### Reproduction sample

https://snack.expo.dev/@idris-timeleft/f615a2

### Relevant log output

```shell
index.js:12 Uncaught Error: findNodeHandle is not supported on web. Use the ref property on the component instead.
    at findNodeHandle (index.js:12:9)
    at useScrollHandler.web.js:28:35
    at commitHookEffectListMount (react-dom.development.js:23189:26)
    at commitPassiveMountOnFiber (react-dom.development.js:24965:13)
    at commitPassiveMountEffects_complete (react-dom.development.js:24930:9)
    at commitPassiveMountEffects_begin (react-dom.development.js:24917:7)
    at commitPassiveMountEffects (react-dom.development.js:24905:3)
    at flushPassiveEffectsImpl (react-dom.development.js:27078:3)
    at flushPassiveEffects (react-dom.development.js:27023:14)
    at react-dom.development.js:26808:9
```

idrisssakhi:
issue is related to "react-native-web": "0.20.0",

erickreutz:
Running into this as well. react-native-web@0.20.0 deprecated findNodeHandle (as it should). BottomSheet will need to update it's code not to use it.

ammarfaris:
+1

_In React 19, findNodeHandle has been removed. This function, previously used to access the native DOM node instance from a React component, is no longer supported due to changes in React's internal architecture and its approach to handling the DOM. The removal of findNodeHandle aligns with React's shift towards more modern and efficient ways of interacting with the DOM._

giaset:
+1 -- completely breaks on web if we use BottomSheetScrollView

giaset:
Here is my patch that got it working on web

[@gorhom+bottom-sheet+5.1.2.patch](https://github.com/user-attachments/files/20024191/%40gorhom%2Bbottom-sheet%2B5.1.2.patch)

doubleedev:
Is there a fix for this yet ?

"@gorhom/bottom-sheet": "^5.1.2",

[...]

PR #2269, opened 2025-05-03
fix(#2237): remove findNodeHandle as it has been removed

TNAJanssen:
## Motivation

See bug #2237 

findNodeHandle is deprecated in newer react-native versions, and is removed from react-native-web 0.20.0 so it needs to be removed from this code as well.

I have tested this locally, on iOS and on Web. Would love to get feedback on the solution, if it's the correct way to do it or not. Also cleaned the code by removing: src/utilities/getRefNativeTag.web.ts

### DIFF
--- a/src/hooks/useScrollHandler.web.ts
+++ b/src/hooks/useScrollHandler.web.ts
@@ -1,5 +1,4 @@
 import { type TouchEvent, useEffect, useRef } from 'react';
-import { findNodeHandle } from 'react-native';
 import { useSharedValue } from 'react-native-reanimated';
 import { ANIMATION_STATE, SCROLLABLE_STATE } from '../constants';
 import type { Scrollable, ScrollableEvent } from '../types';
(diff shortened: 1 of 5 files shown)


erickreutz:
Why did the lock file change so much?

TNAJanssen:
Because i used NPM, i will revert that file tomorrow.

gorhom:
Thanks @TNAJanssen for submitting this PR, it addresses the issue for ScrollView, but not FlatList.

Screenshot 2025-05-04 at 18 01 41

TNAJanssen:
@gorhom thanks for the feedback, i will pick it up Tuesday as it's a day off tomorrow.

gorhom:
thanks for your contribution 🙌, i fixed it in https://github.com/gorhom/react-native-bottom-sheet/commit/47a95f517ab5b4680d0f5a45b09464911aafd35e 

i shall release it shortly

TNAJanssen:
> thanks for your contribution 🙌, i fixed it in https://github.com/gorhom/react-native-bottom-sheet/commit/47a95f517ab5b4680d0f5a45b09464911aafd35e 
> 
> 
> 
> i shall release it shortly 

I think you made one mistake, you're returning the result of the same function instead of _findNodeHandle.

gorhom:
good catch @TNAJanssen 🤦‍♂️ .. published a new patch release https://github.com/gorhom/react-native-bottom-sheet/releases/tag/v5.1.4

Issue #2255, opened 2025-04-25
[Bug]: keyboardBehavior="interactive" causes the bottom sheet to expand further than maxDynamicContentSize


souyahia:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS

### What happened?

When a bottom sheet has dynamic sizing and keyboard behavior set to `interactive`, if the bottom sheet is already fully expanded, when opening the keyboard, the bottom sheet will expand further than the `maxDynamicContentSize` should allow to. Also, this causes to have the `onAnimate` event fired with a `toIndex` value equal to `-1`, even though the bottom sheet is still open, just not at a known snapIndex. This is annoying because in our app, a lot of our logic is based upon the fact that `index === -1` = the bottom sheet is closed.

### Reproduction steps

Define a BottomSheetModal with :
- a `BottomSheetTextInput`
- `enableDynamicSizing={true}`
- `maxDynamicContentSize` set to a certain value
- `keyboardBehavior` set to `interactive` (default value)

### Reproduction sample

https://snack.expo.dev/@souyahia-mistral/bottom-sheet---issue-reproduction-template

souyahia:
As a temporary solution, I patched the lib by adding the `fromPosition` and `toPosition` parameters to the `onAnimate` callback, so that in my app, during the animation event handlers, I base all of my "isOpening" / "isClosing" logic on the position and not on the index of the animate callback.

I then set a value to the `topInset` prop of the BottomSheetModal to prevent it from expanding further even when the keyboard opens. This works, but it is not ideal because when the user overdrags the modal past the inset, the overflow is hidden and it feels not "polished". But at least it works haha

By the way, I think it would be cool to have the `fromPosition` and `toPosition` params available in the `onAnimate` callback in the lib, I think they are useful.

For those interested, here is the patch :

```
diff --git a/src/components/bottomSheet/BottomSheet.tsx b/src/components/bottomSheet/BottomSheet.tsx
index a6f048e7581eefb63c41cf2a383d9f2561e5e7e3..e2043fb122084cf81f170a38c630715b7974980d 100644
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -603,7 +603,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
     );
[...]

PR #2271, opened 2025-05-05
feat: Add position params in onAnimate event

souyahia:

Add `fromPosition` and `toPosition` params to the `onAnimate` event.

## Motivation
There are certain cases where this could be useful (we already have the position info in the `onChange` callback). For example, when the sheets is animating to a position that is not part of the specified snap points, the `toIndex` will be `-1` (which is expected since the snap point does not exist) even though the sheet will be closed. See [this issue](https://github.com/gorhom/react-native-bottom-sheet/issues/2255) for instance.


gorhom:
thanks @souyahia for this PR, it should be release shortly

PR #2282, opened 2025-05-09
Fix Accessibility on Background Containers

braandl:
In the current state, the screenreader is always focusing containers on its way down to the contents of the bottomsheet.

this is (possibly) not the wanted behavior, so we changed it.


anoopmm:
Updates

braandl:
Hm. This bug is still present :/

daviddamilola:
This is still an issue in 2025

braandl:
Oo maybe it is time for a fork? Is it still maintained?

PR #2284, opened 2025-05-11
refactor: rewrote the layouting logic

gorhom:
## Motivation

The goal of this PR is 1) simplify, 2) flatten, 3) optimise the components layouting. this would enable me to integrate LegendList in a follow up PR.


rklf:
Thanks @gorhom ! Is there a PR for BottomSheetLegendList that we can follow ?

xanderdeseyn:
We would also be very interested in using LegendList here!

PR #2321, opened 2025-06-11
Fix findNodeHandle.web.ts for BottomSheetFlatList

Vadko:
## Motivation

When using BottomSheetFlatList with react-native-web, the web implementation of findNodeHandle fails to locate the scrollable node for FlatList. However, this node can be accessed via the [getScrollableNode](https://github.com/necolas/react-native-web/blob/0.20.0/packages/react-native-web/src/vendor/react-native/FlatList/index.js#L408) method.
This PR adds support for retrieving the scrollable node using getScrollableNode, ensuring proper functionality on the web.

### DIFF
--- a/src/utilities/findNodeHandle.web.ts
+++ b/src/utilities/findNodeHandle.web.ts
@@ -22,6 +22,14 @@ export function findNodeHandle(
     }
   } catch {}
 
+  try {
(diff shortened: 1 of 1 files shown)


Vadko:
cc @gorhom

PR #2328, opened 2025-06-13
feat(android): add bottom tab bar height property

pakerwreah:
## Motivation

Add a new property to fix extra padding below text input on android when the bottom sheet is inside a stack that contains a bottom tab bar.

The provided value usually comes from `@react-navigation/bottom-tabs` hook `useBottomTabBarHeight`.

Before the feature proposed on this PR, I had to dynamically change `bottomInset` based on the keyboard state, which was causing the bottom sheet to randomly close instead of collapsing, and also performance issues due to state change during animation.

## Tests
- Works as expected with and without a bottom tab bar
- Continues to work nice on iOS (we ignore the value)

## Screenshots
Collapsed|Issue (extra padding)|Fixed
-|-|-
collapsed|extra-padding|expected


pakerwreah:
not stale

Issue #807, opened 2022-01-10
[v4]  Adding index={-1} in certain devices blocks the view (Example device: Oneplus 5)


aditya-keri-betterhalf:
# Bug

In certain devices like the [OnePlus 5](https://www.oneplus.in/support/spec/oneplus-5), adding an `index={-1}` is causing the entire screen to be non-functional, i.e, we're unable to click anywhere on the screen. If we remove this prop, then everything seems to be working fine. But of course we need the BottomSheet to be in closed state initially, hence not adding this prop is not a solution.
But this does work fine on other devices we've tested. 

This issue is affecting many users in Production, since we're using BottomSheet in many screens and the Users are unable to use those screens.

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.1.5   |
| react-native                    | 0.64.2  |
| react-native-reanimated         | 2.2.4  |
| react-native-gesture-handler    | 1.10.3  |

## Steps To Reproduce

In certain devices only :
Add an index={-1} in the BottomSheet is causing the entire screen to be non-functional. It is as though there's an overlay on top of the screen which doesn't let us access whatever is currently rendered.

Describe what you expected to happen:

It should not render the screen non-functional.

## Reproducible sample code

```import BottomSheetGorhom, { BottomSheetBackdrop } from '@gorhom/bottom-sheet';
import React, { useCallback } from 'react';
import { StyleSheet, Keyboard } from 'react-native';
import PropTypes from 'prop-types';
import { bottomSheetHeaderHandler } from 'colors';
import {
  heightPercentageToDP as hp,
  widthPercentageToDP as wp,
} from 'react-native-responsive-screen';

const renderBackdrop = props => (
  <BottomSheetBackdrop {...props} disappearsOnIndex={-1} appearsOnIndex={0} />
);

const BottomSheet = React.forwardRef((props, ref) => {
  const { children, onChange, ...restProps } = props;

  const handleChange = useCallback(index => {
    if (index === -1) {
      // When the BottomSheet is closed, dismiss any open Keyboard
      Keyboard.dismiss();
    }
    // Call the actual onChange handler passing in the index
    onChange?.(index);
  }, []);

  return (
    <BottomSheetGorhom
      index={-1}
      enablePanDownToClose
      backdropComponent={renderBackdrop}
      handleIndicatorStyle={styles.handleIndicatorStyle}
      ref={ref}
[...]

Issue #1882, opened 2024-07-11
[v4] | BottomSheetModal is not working on some Android phones


MarchewkaMatthew:
# Bug

As mentioned in this automatically closed issue: https://github.com/gorhom/react-native-bottom-sheet/issues/1819. On some Android phones bottom sheet modal is not visible. For now, 3 users reported that issue. List of their devices:
- Xiaomi Mi 9T (Android 13)
- Xiaomi Mi 9T Pro
- Samsung A72

We have tried to reproduce this issue on Browserstack (using Xiaomi Note 8 Android 9), but everything worked correctly. 

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4.5.1   |
| react-native                    | 0.74.2  |
| react-native-reanimated         | 3.10.1   |
| react-native-gesture-handler    | 2.16.2   |

## Steps To Reproduce

1. Create the basic bottom sheet modal with visible variable (initial value false)
2. Change the value to "true"
3. Modal is not visible - nothing changes in the app

## Describe what you expected to happen:

The bottom sheet modal is visible on all Android phones.

## Reproducible sample code

Code for the Bottom sheet modal that we are using:

```
    <BottomSheetModal
      ref={bottomSheetModalRef}
      handleComponent={null}
      onDismiss={onClose}
      onChange={handleSheetPositionChange}
      backdropComponent={props => (
        <BottomSheetBackdrop {...props} opacity={0.2} disappearsOnIndex={-1} appearsOnIndex={0} />
      )}
      backgroundStyle={{ borderRadius: 24 }}
      visible={visible}
      onClose={onClose}
      snapPoints={[410]}
    />
```

OFD16:
same problem with me. I have Xiaomi red mi note 10 5G (Android 13). And bottomsheets was working after i got apk build, bottomsheets does not working on my device too. So i change the package i am using [react-native-actions-sheet](https://github.com/ammarahm-ed/react-native-actions-sheet) its working but now its not working too in release(apk) mode ([issue](https://github.com/ammarahm-ed/react-native-actions-sheet/issues/381)). Also [react-native-modals](https://www.npmjs.com/package/react-native-modals) package not working on release (apk) mode too.

I have already enabled `android > gradle.properties`
`newArchEnabled=true`. I tihnk this is causing the packages problems. It can be general problem i think.

MarchewkaMatthew:
[...]

Issue #566, opened 2021-08-07
Nothing works on the screen  on some android devices if I will try to use this component.


beqramo:
# Bug
Hi,
I  use BottomSheetBackdrop component and on some devices when the screen is mounted, no button works on that screen, it feels like something is overlayed on top of that. This bug only exists on android's some devices.  
For example, it doesn't work: on Huawei mate 8 version 7.0,  Nokia 2.1 version 10, Samsung galaxy s10 version 10, techno camon 15 version 10. 

On the other hand, I tested it on my phones: Meizu note 5 version 7.0, Google pixel 3 version 10 and everything works perfectly( as I said even IOS doesn't have such a problem ). 

Thanks.
## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet         | ^3.6.6   |
| react-native                    | 0.64.1   |
| react-native-reanimated         | ^2.2.0   |
| react-native-gesture-handler    | ^1.10.3   |

## Steps To Reproduce

1. Run codes
2. Try to touch anything on the screen when you use this library

Describe what you expected to happen:

1. screen should be clickable.

## Reproducible sample code
```jsx
function ScreenComponent() {
  const bottomSheetRef = useRef<BottomSheet>(null);
  const [isInsightActive, setIsInsightActive] = useState(true);

  const { bottom } = useSafeAreaInsets();

  const onIsItHelpfulPress = useCallback(() => {
    bottomSheetRef.current?.expand();
  }, []);

  return (
    <Screen stretch>
      <Header withTitle />
      <Row justifySpaceBetween alignCenter px={7} mt={7}>
        <Touchable active={isInsightActive} onPress={onIsItHelpfulPress}>
          <Text fontSize={4} fontWeight={3}>
            Insights
          </Text>
        </Touchable>
      </Row>

      <BottomSheet
        ref={bottomSheetRef}
        index={-1}
        snapPoints={[0, 200]}
        backdropComponent={BottomSheetBackdrop}
      >
        <Column mt={7} stretch paddingBottom={`${bottom}px`}>
          <Text fontSize={5} fontWeight={3} px={7}>
            Is this helpful?
          </Text>
        </Column>
      </BottomSheet>
    </Screen>
  );
}
```

harsh-siriah:
I was facing a similar issue. 
My workaround was to put 0 as the first element in my snap points array and then pass index={0} in the <BottomSheet> component.

[...]

Issue #527, opened 2021-07-11
BottomSheetBackdrop component doesn't show up at all when setting the `appearsOnIndex` prop


ramirezabel:
# Bug

The `BottomSheetBackdrop` component wont show up when setting the `appearsOnIndex={0}` for example. As I understand the prop is to tell the backdrop to show up when the snap point in on the index provided. If thats not the case then can someone provide a way of doing that. In my case the backdrop doesn't show when the index is at `0` only when dragging the component up to the other snap points does it show, however like mentioned above, if I set the property of the backdrop `appearsOnIndex={0}` it doesn't show at all. 

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet         | 4.0.0-alpha.19  |
| react-native                    |  ^0.64.1   |
| react-native-reanimated         |  ^2.1.0  |
| react-native-gesture-handler    |  ^1.10.3   |

## Steps To Reproduce

1.  Set the `BottomSheetBackdrop` prop `appearsOnIndex={0}` 
2. Open the bottom sheet modal and the backdrop doesn't show

Describe what you expected to happen:

1. Set the `BottomSheetBackdrop` prop `appearsOnIndex={0}` 
2. Backdrop should show up as soon as the modal opens to the first snap point

## Reproducible sample code

<img width="810" alt="Screen Shot 2021-07-10 at 8 19 11 PM" src="https://user-images.githubusercontent.com/10616358/125179367-19ed1080-e1bc-11eb-8348-2a59a28bb5d9.png">

<img width="605" alt="Screen Shot 2021-07-10 at 8 17 51 PM" src="https://user-images.githubusercontent.com/10616358/125179351-f924bb00-e1bb-11eb-8d37-a9f6331d9027.png">

gorhom:
@ramirezabel  you need to provide `disappearsOnIndex` with `-1` , the default value is `0`

https://github.com/gorhom/react-native-bottom-sheet/blob/73decd3f13f17f585796556fc2bd43f408be8603/src/components/bottomSheetBackdrop/types.d.ts#L24-L29

pfcodes:
> @ramirezabel you need to provide `disappearsOnIndex` with `-1` , the default value is `0`
> 
> https://github.com/gorhom/react-native-bottom-sheet/blob/73decd3f13f17f585796556fc2bd43f408be8603/src/components/bottomSheetBackdrop/types.d.ts#L24-L29

[...]

PR #2342, opened 2025-06-20
fix: Math.floor animatedIndex value in BottomSheetBackdrop to fix And…

margreen94:
## Motivation
On some Android devices, with disabled animations, any screen with the <GorhomBottomSheet/> component appears to "freeze". No component on the screen is able to be interacted with. The user cannot open the bottom sheet, click any buttons, or even scroll.

## Steps to reproduce 
1. Use an Android device (I use a Pixel 9, Android 16 simulator)
2. Go into the device settings and disable animations (Settings > Accessibility > Color & motion > toggle Remove animations "on")
3. Go to a screen on your device that includes a <GorhomBottomSheet/> component that is rendered "closed" on mount with prop: index={-1} and has a `backdropComponent`. 
4. Actual: User will not be able to interact with any component on screen. Expected: User should be able to interact with the app like normal and open the bottom sheet.

## Problem
The root of the issue appears to be the `useAnimatedReaction` hook in the `BottomSheetBackdrop.tsx` component. The screen appears to "freeze" because although the backdrop is not visible, the backdrop logic is incorrectly disabling the touchability of the screen. On certain Android devices with reduced animations "on", the sharedValue of `animatedIndex.value` only reaches the value of -0.999… and never crosses the threshold of -1 which is the `disappearsOnIndex` value. Thus, the return value on line 118 will always equal "false" and cause the `handleContainerTouchability` function to disable the pointerEvents and appear to "freeze" the screen.

## Solution
The proposed solution is to Math.floor() the `animatedIndex` sharedValue before comparing it to the `disappearsOnIndex` value. This will ensure that if the sharedValue does not quite reach the `disappearsOnIndex` value, it will not incorrectly disable touchability. In my app, I have this fixed as a local patch but am looking for a more long-term solution with this PR.

Listed below are previous closed issues related to this fix:
https://github.com/gorhom/react-native-bottom-sheet/issues/807
https://github.com/gorhom/react-native-bottom-sheet/issues/1882
https://github.com/gorhom/react-native-bottom-sheet/issues/527
https://github.com/gorhom/react-native-bottom-sheet/issues/566

### DIFF
--- a/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
+++ b/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
@@ -115,7 +115,7 @@ const BottomSheetBackdropComponent = ({
 
   //#region effects
   useAnimatedReaction(
-    () => animatedIndex.value <= disappearsOnIndex,
+    () => Math.floor(animatedIndex.value) <= disappearsOnIndex,
     (shouldDisableTouchability, previous) => {
       if (shouldDisableTouchability === previous) {
         return;


aarontuftss:
Confirmed this fixed issues on my Pixel 9a making device unresponsive. 

Inspecting the elements showed the BottomSheetBackdrop was covering the entire screen, consuming pointer events.

The issue would occur randomly with animations turned on. Turning animations off however made the issue appear 100% of the time. 

----

Although Math.floor() fixes the blocked pointer events, I noticed it introduced a new behavior where you can tap through the backdrop until the bottom sheet is completely done opening.

~~I am wondering if Math.round() would be a better solution without implementing any complex logic?~~

Both Math.floor() & Math.round() leave too big of a window where the backdrop can still be touched through. If you need the bottom sheet to restrict page interaction, that can be a serious issue.

I do not think this package is inherently doing anything wrong & the bug is at a much lower level since its so device & OS specific.

Even though its a 'magic number' solution, I will probably use this workaround until the root issue is solved.

```
  const EPSILON = 0.001

  useAnimatedReaction(
    () => animatedIndex.value <= (disappearsOnIndex + EPSILON),
    (shouldDisableTouchability, previous) => {
      if (shouldDisableTouchability === previous) {
        return;
      }
      runOnJS(handleContainerTouchability)(shouldDisableTouchability);
    },
    [disappearsOnIndex]
  );
```

gorhom:
thanks @margreen94 for submitting this PR and detailing the issue. 

For me based on the information provided, it seems to me that we need to address the issue with no-animation accessibility issue on Android devices to be handled properly. Unfortunately accessibility is not my strong skill to solve this issue. Let me investigate it first, then i'll decide either it worth investing the root cause or accept the short term solution.

jslok:
Seeing this issue on rn 0.81.4 with new arch but noting this is with reanimated v4 which I realize is not listed as supported by bottom sheet. Several users reported the issue on Android with various OS versions and different devices. Not sure what is the common denominator but I was able to reproduce it with a Pixel 9 sdk 36 emulator.

I did not see this issue on 0.75.5 old arch with reanimated v3.

The PR does solve the issue for me. Thanks!

antek-p:
Hello, is it going to be merged? im facing freezing issue on android device using expo sdk53, reanimated v4

batiiy:
+1
"react-native": "0.81.4",
"react-native-reanimated": "^3.19.1",

margreen94:
@gorhom Hello again! Any update on this being merged?

Nikita0x:
I encountered this issue and it was quite unpleasant. Works on Web and iOS but on Android whole UI becomes unresponsive.
But without it - there is no "backdrop" component - so you *have* to use it - leaving no chance 

Thank you for this PR @margreen94! 
Did you confirm that your PR fixes this freeze of the UI on androids? 
if `Yes`  - then why it is not merged already?

MateoParodi:
+1 This NEEDS to be merged. for now with a patch we are good. but it affected lots of our Android users having the app "frozen"

0xPT:
+1 blocking any widely adopted app from using this repository

kr-yeon:
+1

margreen94:
@gorhom Any update on when this will be merged?

Hristiyyann:
+1 facing heavily this issue

evparth:
In our case, this issue is present in specific phones/Android OS versions, for example a Xiaomi Mi 9T with Android 11 (Remove animations android setting independent).
I can confirm the PR fixes the issue, but currently with a patch. Would be nice to include this PR in upcoming versions.

calvinkei:
Confirm this issue happens on Samsung S23 Ultra and some Xiaomi phones with v5.2.8. This PR fixed it

antek-p:
Switched to react-native-true-sheet and it's working better.

felixchen-wordup:
Confirm this issue happens on Pixel 10 with multi user mode (v5.2.13).
This PR save my life.

adokuru:
@gorhom Please can this be merged super important

PR #2292, opened 2025-05-22
support SectionList from findNodeHandle.web.ts

btoo:
## Motivation

fixes https://github.com/gorhom/react-native-bottom-sheet/issues/2237#issuecomment-2900104043 for react-native-web 20

PR #2359, opened 2025-07-13
try getScrollableNode in findNodeHandle.web.ts

btoo:
## Motivation

a follow up to https://github.com/gorhom/react-native-bottom-sheet/pull/2292#issuecomment-2910535509, i realized that i'm still getting an issue on `BottomSheetSectionList` (and maybe others?) on web like

image

### DIFF
--- a/src/utilities/findNodeHandle.web.ts
+++ b/src/utilities/findNodeHandle.web.ts
@@ -22,6 +22,14 @@ export function findNodeHandle(
     }
   } catch {}
 
+  try {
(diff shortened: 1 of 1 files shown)


robertkossendey:
This is blocking us as well. We pnpm patched the module for now, but it would be lovely if this could get merged.

gorhom:
@btoo  @robertkossendey if you could provide me more details on the issue, maybe repo sample code then i can investigate it

PR #2387, opened 2025-07-29
Disable overshootClamping for small bottom sheets to avoid Reanimated issue

sharifhh:
Hi! 

Reanimated v4 introduced a bug in withSpring
https://github.com/software-mansion/react-native-reanimated/issues/7947
https://github.com/software-mansion/react-native-reanimated/pull/7803

this is a temp solution to make this library work for small bottom sheets

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -596,10 +596,15 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
           isAnimatedOnMount.value = true;
         }
 
+        // reset values
         animatedAnimationSource.value = ANIMATION_SOURCE.NONE;
         animatedAnimationState.value = ANIMATION_STATE.STOPPED;
       },
-      []
+      [
+        animatedAnimationSource,
+        animatedAnimationState,
+        isAnimatedOnMount,
+      ]
     );
     const handleNewAnimatedValues = useCallback(
       function handleNewAnimatedValues() {
@@ -624,7 +629,12 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
         animatedNextPositionIndex.value = INITIAL_VALUE;
         animatedContainerHeightDidChange.value = false;
       },
-      []
+      [
+        animatedContainerHeightDidChange,
+        animatedNextPosition,
+        animatedNextPositionIndex,
+        isForcedClosing,
+      ]
     );
     const animateToPosition: AnimateToPositionType = useCallback(
       function animateToPosition(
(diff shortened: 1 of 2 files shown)


matiaswastaken:
With this fix it partially works, but the starting animation is gone meaning when you open the sheet you arent getting the smooth animation (basically same as animateOnMount=false)

sharifhh:
actually my whole fix is not needed, the issue is in the new react-native-reanimated withSpring, try to play with the animation configuration and you will see that it works

matiaswastaken:
@sharifhh oh hmm, can you share what steps i have to do get this working?

sharifhh:
these are the default configs

```
const ANIMATION_EASING: Animated.EasingFunction = Easing.out(Easing.exp);
const ANIMATION_DURATION = 250;

const ANIMATION_CONFIGS_IOS = {
  damping: 500,
  stiffness: 1000,
  mass: 3,
  overshootClamping: true,
  restDisplacementThreshold: 10,
  restSpeedThreshold: 10,
};

const ANIMATION_CONFIGS_ANDROID = {
  duration: ANIMATION_DURATION,
  easing: ANIMATION_EASING,
};

const ANIMATION_CONFIGS =
  Platform.OS === 'ios' ? ANIMATION_CONFIGS_IOS : ANIMATION_CONFIGS_ANDROID;
  ```
  
  
   now 
   ```
     restDisplacementThreshold: 10,
  restSpeedThreshold: 10,
   ```
   are removed from withSpring so ignore them, and just set overshootClamping as false and it will be resolved.
   
   I found the issue in reanimated and starting a thread there, will keep you posted and link it here

**Comment by sharifhh** [comment]
https://github.com/software-mansion/react-native-reanimated/issues/7947

**Comment by matiaswastaken** [comment]
Thanks for the effort. Is there any quickfix patch that can be applied while we wait, I would need this to be fixed ASAP.

EDIT: Ah just set overshootClamping: false and it seems to work fine. Thanks.

**Comment by sharifhh** [comment]
https://github.com/software-mansion/react-native-reanimated/pull/7948

**Comment by Willham12** [comment]
Perfect, thanks mate. fix #2272

**Comment by justynpollard1** [comment]
Thanks, setting overshootClamping: false in animation config fixed this for me as well. Before my sheets would blink quickly and not appear or just not appear at all.

**Comment by KatChaotic** [comment]
Thanks, I apply this changes locally with patch and now modals start working on iOS after migration to new versions

**Comment by Yurii-Lutsyk** [comment]
Thanks you!!!

**Comment by Splanis** [comment]
OMG thanks

**Comment by gorhom** [comment]
thanks @sharifhh for submitting this PR, unfortunately these days im really busy with my daily job and did not see this PR in time.

looking at the issue and it seems that is coming from reanimated, and they have now fixed it [[link](https://github.com/software-mansion/react-native-reanimated/pull/7803)].

**Comment by santhosh-umapathi-wikifolio** [comment]
This works. Thanks all 

```
animationConfigs={{
        overshootClamping: false,
      }}
```

PR #2399, opened 2025-08-07
Add accessibilityViewIsModal prop to bottomsheetbackdrop

tglasgow-7f:
We ran into an issue with a11y when managing keyboard focus. Having a configurable accessibilityViewIsModal prop gives us more control over how we manage a11y throughout our app. Without it, the backdrop takes keyboard focus and doesn't allow the contents of the bottom sheet to take focus.


sergiogirado:
Can we also support `role` and `accessibilityLabelledBy`?
1. [role](https://reactnative.dev/docs/accessibility#role) can be set to `dialog` (and this takes precedence over `accessibilityRole`)
2. [accessibilityLabelledBy](https://reactnative.dev/docs/accessibility#accessibilitylabelledby-android) is usefull for scenarios were the bottom sheet has a title element we can bind. Docs says works on android only but this is handy in testing library as allows to do a clean `getByRole('dialog', { name: 'label' })`.

sergiogirado:
> Can we also support `role` and `accessibilityLabelledBy`?
> 
> 1. [role](https://reactnative.dev/docs/accessibility#role) can be set to `dialog` (and this takes precedence over `accessibilityRole`)
> 2. [accessibilityLabelledBy](https://reactnative.dev/docs/accessibility#accessibilitylabelledby-android) is usefull for scenarios were the bottom sheet has a title element we can bind. Docs says works on android only but this is handy in testing library as allows to do a clean `getByRole('dialog', { name: 'label' })`.

I ended up opening another PR #2432

PR #2420, opened 2025-08-20
fix: conditionally apply web-only cursor style to avoid TypeScript error

kirstilynn:
## 🧠 Context

This component previously used:

```ts
// @ts-ignore supported on web
cursor: 'grab',
```
to support cursor styling on the web platform. However, this still triggers a TypeScript error when using `"jsx": "react-native"` or when TypeScript re-validates `style` against `ViewStyle` — because the inferred object type includes `cursor`, which isn't allowed on native.

Even with `"skipLibCheck": true`, the error shows up **at usage**, not declaration, so `@ts-ignore` isn't enough.

---

## ✅ What’s fixed

The style is now wrapped in a platform check:

```ts
container: {
  ...(Platform.OS === 'web'
    ? { padding: 10, cursor: 'grab' }
    : { padding: 10 }),
},
```

This ensures the `cursor` property only exists at runtime when `Platform.OS === 'web'`, so it's no longer included in the type — and TS no longer complains when passing it to a `<View style={...}>`.

---

## 📌 Why this matters

- Removes the need for `@ts-ignore`
- Fixes valid TS errors that still happen despite the ignore
- Maintains behavior for web with zero runtime impact elsewhere

---

Let me know if you'd like to validate this further or want additional changes.

### DIFF
--- a/src/components/bottomSheetHandle/styles.ts
+++ b/src/components/bottomSheetHandle/styles.ts
@@ -1,11 +1,11 @@
-import { StyleSheet } from 'react-native';
+import { Platform, StyleSheet } from 'react-native';
 import { WINDOW_WIDTH } from '../../constants';
 
 export const styles = StyleSheet.create({
   container: {
-    padding: 10,
-    // @ts-ignore supported on web
-    cursor: 'grab',
+    ...(Platform.OS === 'web'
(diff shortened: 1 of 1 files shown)


gorhom:
thanks for @kirstilynn for submitting this PR 🙌

PR #2457, opened 2025-09-05
fix: set keyboard height to 0 on TextInput blur

goknsh:
This fixes an issue where the bottom sheet does not restore down to the bottom of the screen when the input blurs and the software keyboard is hidden.

Ensure that the modal's height + software keyboard height is greater than the screen height. Otherwise you won't experience the bug.

Issue reproduction: https://github.com/goknsh/react-native-bottom-sheet-repro


goknsh:
not stale

goknsh:
not stale

PR #2481, opened 2025-09-15
fix: ts compile error on Animated has no exported member EasingFunction

LeunensMichiel:
With the latest react-native-reanimated version, and expo sdk 54 (new arch), I got a ts compile error from this library

tsc --build --noEmit

error TS2694: Namespace '"/Users/michielleunens/Documents/code/allride_app/node_modules/react-native-reanimated/lib/typescript/Animated"' has no exported member 'EasingFunction'

I've updated the import locally like this and I was able to compile again

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -68,7 +68,7 @@ enum SNAP_POINT_TYPE {
   DYNAMIC = 1,
 }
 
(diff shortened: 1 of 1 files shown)


Aduomas (approved):

LeunensMichiel:
not stale

LeunensMichiel:
Fixed on main, closing

PR #2486, opened 2025-09-18
Update constants.ts

joshua-zbni:
## Motivation

This PR fixes a TypeScript typing issue in `react-native-reanimated` where the `EasingFunction` type is being imported but not exported from the `Animated` namespace.  

Currently, this results in the following error during compilation:

```
Namespace '.../node_modules/react-native-reanimated/lib/typescript/Animated'
has no exported member 'EasingFunction'.
```

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -69,7 +69,7 @@ enum SNAP_POINT_TYPE {
   DYNAMIC = 1,
 }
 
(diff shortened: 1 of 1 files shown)


AnkushSarkar10:
@gorhom can this get merged?

gorhom:
thanks for submitting the PR , it should be released shortly

Issue #2509, opened 2025-10-01
[Bug]: Bottom sheet does not go down after closing the keyboard if you use BottomSheetScrollView


exzos28:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS

### What happened?

[The issue](https://github.com/gorhom/react-native-bottom-sheet/issues/2460) still exists, everything is the same.

Problem: When using `BottomSheetScrollView`, if you open and close the keyboard, `BottomSheetScrollView` starts to occupy all the space where the keyboard was and does not return to its original position.
This only happens if the scroll content overflows when the keyboard is open (see the two videos below that demonstrate this).

### Reproduction steps

https://github.com/user-attachments/assets/35e00184-67ca-4e99-b363-bcfca67281aa

https://github.com/user-attachments/assets/ec6e0720-eaaf-49f0-97ee-29661cfe2267

### Reproduction sample

https://snack.expo.dev/@exzos28/f615a2?platform=ios

ChaparroAlberto:
I have the exact same error, I will continue to be attentive if you find a solution.

exzos28:
@ChaparroAlberto I temporarily fixed this like this:
```
    const isKeyboardVisible = useKeyboardState(state => state.isVisible); // or Keyboard.addListener('keyboardDidHide')
    useEffect(() => {
      if (Platform.OS !== “ios”) {
        return;
      }
      if (!isKeyboardVisible) {
        bottomSheetRef.current?.snapToIndex(0);
      }
    }, [bottomSheetRef, isKeyboardVisible]);
```

The call happens with some delay, but it's better than nothing.

ChaparroAlberto:
Okay, thanks a lot. Do you put that code in the component where you use it? Or where do you put that code snippet?

exzos28:
in the place where you have access to the modal ref

MuhammadAnsoriNasution:
> [@ChaparroAlberto](https://github.com/ChaparroAlberto) I temporarily fixed this like this:
> 
> ```
>     const isKeyboardVisible = useKeyboardState(state => state.isVisible); // or Keyboard.addListener('keyboardDidHide')
>     useEffect(() => {
>       if (Platform.OS !== “ios”) {
>         return;
>       }
>       if (!isKeyboardVisible) {
>         bottomSheetRef.current?.snapToIndex(0);
>       }
[...]

Issue #2465, opened 2025-09-09
[Bug]: keyboardBlurBehavior={'restore'} not working


florian-deroo:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

Android, iOS

### What happened?

keyboardBlurBehavior={'restore'} isn't working, the position of the modal does not restore when i close the keyboard and there is a topInset

### Reproduction steps

- Open input
- Close input

### Reproduction sample

https://snack.expo.dev/@florianfrdev/bottom-sheet---issue-reproduction-template?platform=ios

### Relevant log output

```shell

```

doguyilmaz:
```tsx
<BottomSheetModal
    ref={bottomSheetModalRef}
    snapPoints={["60%", "100%"]}
    enableDynamicSizing={false}
    topInset={100}
    keyboardBlurBehavior={'restore'} 
>
```

actually adding "100%" to snapPoints helped in my case and your snack, yeah probably wont fit all the cases but could solve your issue temporarily.

[your sample with the fix](https://snack.expo.dev/@doguyilmaz/f615a2)

krishnan-456:
yes, same issue here with latest version

florian-deroo:
@gorhom

krishnan-456:
does anyone find solution for this or do we have any hack to temporary fix this ?

Akyna:
The same issue

```
"expo": "~53.0.22",
"react-native": "0.79.5",
"@gorhom/bottom-sheet": "^5.2.6"
```

krishnan-456:
this is happening only when enableDynamiceSize={true}

krishnan-456:
does anyone faced this issue with IOS ?

mefjuu:
@krishnan-456 Yup, this is flawed on Android as well.

Akyna:
We downgraded to the previous version (5.1.2) - works like a charm.
```
"expo": "~53.0.22",
"react-native": "0.79.5",
"@gorhom/bottom-sheet": "5.1.2"
```

krishnan-456:
@Akyna can you share the reanimated version ? because it's throwing error.

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/b15d2043-973a-4ab8-9314-67d88b09a7d5" />

Akyna:
> [@Akyna](https://github.com/Akyna) can you share the reanimated version ? because it's throwing error.




[...]

Issue #2460, opened 2025-09-08
[Bug]: Bottom sheet does not go down after closing the keyboard if you use `BottomSheetScrollView`


DmytroMelnyk26:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

Android, iOS

### What happened?

Bottom sheet does not go down after closing the keyboard if you use `BottomSheetScrollView`

### Reproduction steps

1. open bottom sheet with `BottomSheetScrollView`
2. press on Input
3. press on any other place
4. bottom sheet steel in a center of screen

https://github.com/user-attachments/assets/bb328274-a5bf-4214-bfb3-8ec2f675f739 

### Reproduction sample

https://snack.expo.dev/@dmytromelnykakveo/bottom-sheet---issue-reproduction-template?platform=ios

### Relevant log output

```shell

```

KatChaotic:
Got same issue, temporary fix it on our side using `react-native-keyboard-controller`.
That is not a good solution, little bit laggy, but much better than floating modal
```
const isKeyboardVisible = useKeyboardState(state => state.isVisible);
    useEffect(() => {
        if (isOpen && !isKeyboardVisible && bottomSheetRef.current) {
            // After setting it once, modal starts working fine
            bottomSheetRef.current.snapToIndex(0);
        }
    }, [ bottomSheetRef, isKeyboardVisible ]);
```

adam-sajko:
I found this in the docs:
```tsx
<BottomSheet keyboardBlurBehavior="restore" />
```
https://gorhom.dev/react-native-bottom-sheet/props#keyboardblurbehavior

The default is "none"

DmytroMelnyk26:
@adam-sajko thank you. It helped me

SaqoTadevosyan:
Problem still exist for `<BottomSheetModal />` `keyboardBlurBehavior="restore" ` dosnt help

SergeyMelnyk:
@adam-sajko 
Problem still exist even with keyboardBlurBehavior="restore"

PR #2511, opened 2025-10-03
fix: restore bottom sheet when keyboard is dismissed

michael-pingo-ai:

## Motivation

This PR fixes a bug where the bottom sheet could assign a raw position value to index when the keyboard was shown. This caused animatedCurrentIndex to be set to invalid values (e.g. `235`), which led to broken state and the sheet height not resetting when the keyboard was dismissed.

The fix ensures index is resolved against the detents array, falling back to DEFAULT_KEYBOARD_INDEX if no match is found.

Solves:
#2509 
#2465 
#2460

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -659,7 +659,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
           status === KEYBOARD_STATUS.SHOWN &&
           position !== closedDetentPosition
         ) {
-          index = highestDetentPosition ?? DEFAULT_KEYBOARD_INDEX;
+          index = detents?.indexOf(highestDetentPosition ?? 0) ?? DEFAULT_KEYBOARD_INDEX;
         }
 
         /**


src/components/bottomSheet/BottomSheet.tsx
```diff
@@ -659,7 +659,7 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
           status === KEYBOARD_STATUS.SHOWN &&
           position !== closedDetentPosition
         ) {
-          index = highestDetentPosition ?? DEFAULT_KEYBOARD_INDEX;
+          index = detents?.indexOf(highestDetentPosition ?? 0) ?? DEFAULT_KEYBOARD_INDEX;
```
amillez, line 662: Just found the same bug and was about to open the same pr😁

I think it is better to just get the last item of the array since we know they are sorted in ascending order, right?

```suggestion
          index = detents?.length ? detents.length - 1 : DEFAULT_KEYBOARD_INDEX;
```

itsmitul (approved):

HassanKhan96:
Why was this PR not merged?

src/components/bottomSheet/BottomSheet.tsx
tvillafane: this seems to do the trick!

tsvetlin:
This PR fixed the issue. I've patched the package and it's working correctly now! Thank you @AgustinMJ

seba9999:
Please re-open ? 😞

PR #2519, opened 2025-10-08
Drop doc line mentioning `stackBehavior` is only available in v3

fabriziocucci:
## Motivation

`stackBehavior` should be usable also in v5.

### DIFF
--- a/website/docs/modal/props.md
+++ b/website/docs/modal/props.md
@@ -20,8 +20,6 @@ Modal name to help identify the modal for later on.
 
 ### stackBehavior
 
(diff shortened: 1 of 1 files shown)


flexbox:
Hey @fabriziocucci 
Do you have more context about your request and a reproducible example?

fabriziocucci:
@flexbox the doc seems to indicate that the prop `stackBehavior` is only available in v3 of the library but that doesn't seem to be the case, i.e. the prop is indeed available also for the latest v5 of the library, right?

PR #2569, opened 2025-11-13
Fix example snapPoints format in types definition

pvinis:
Not sure why this was flipped, but heres a quick fix.


keeandev:
@pvinis AI was confused or a developer from our team because of this!

pvinis:
@gorhom any chance to quickly merge this?

PR #2571, opened 2025-11-14
feat: add one snap point change per one swipe

IslamRustamov:
## Motivation

Application I'm working on has a bottom sheet with 3 snap points.

Recently I was informed that there is an issue with it:

1. Bottom sheet is at the last snap point;
2. User quickly swiped a little bit;
3. Bottom sheet goes from last snap point to first one, ignoring the middle snap point.

So I was tasked to solve this by allowing user only one snap point change per swipe, but the library doesn't support this feature. 

Here's the example of the problem:

https://github.com/user-attachments/assets/9bc9a44e-10ac-4c27-83df-9775439062c1

We can see that most of the time, a little swipe is causing our bottom sheet to skip the middle snap point.

Here's the example of how it works with my changes:

https://github.com/user-attachments/assets/d86e3c70-271e-40e3-85f1-bfa66374f320

We can see that now I can't skip the middle snap point and am forced to do one more swipe to get to the next snap point.


IslamRustamov:
~~after I synced my fork - some problems jumped out, going to put in draft until I fix them~~

~~It stopped working the right way after these **detents**~~

~~the only problem that left is that animation is instant in case if I overdrag, will look into that but will be glad for help~~

upd: nervermind, this problem is existing in master, not the problem caused by my changes

https://github.com/user-attachments/assets/b9383977-b491-4494-92ea-c662cb6110ba

Issue #2582, opened 2025-12-04
[Bug]: Disabling animateOnMount - the animation still plays (+ visual jumps when it doesn't)


kliakhovskii-brex:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS, Android

### What happened?

This is somewhat 2 distinct issues, but they seem to be related overall due to initial state of things with `animateOnMount` and a few other things, therefore reporting them as one for now.

tldr: We use bottom sheet inside of tabs, so each tab has a bottom sheet rendered, and therefore mount their own version, some tabs do not let bottom sheet to be collapsed state. Because it's predictable environment, we always know the height and do not require the dynamic sizing.

After `5.1.6` -> `5.2.6` upgrade surfaced two issues:
- `animateOnMount={false}` stops working when both height provided and handle is supplied as null.
- `animateOnMount={false}` (when it works) renders the initial frame in collapsed state, then jumps into position

### Reproduction steps

# Issue #1:
Description - supplying both `containerLayoutState.height` and `handleComponent={null}` breaks `animateOnMount={false}`

## Steps:
The following code would start playing animation if `handleComponent={null}` is uncommented
```
const containerLayoutState = useSharedValue<ContainerLayoutState>({
    height: screenHeight,
    offset: {
      top: 0,
      left: 0,
      right: 0,
      bottom: 0,
    },
  });

...

<BottomSheet
        style={{ flex: 1 }}
        index={1}
        snapPoints={[104, 729, 792]}
        enableDynamicSizing={false}
        // handleComponent={null} // <- uncomment this line to see the animation start playing agoing.
        containerLayoutState={containerLayoutState}
        animateOnMount={false}
      >

```

## Quick investigation:
I have looked at internals for quick investigation, and funny enough, this is very straightforward issue:
- We get [this effect](https://github.com/gorhom/react-native-bottom-sheet/blob/v5.2.7/src/components/bottomSheet/BottomSheet.tsx#L1767-L1774) executed with original `_providedIndex` value. And because the `animateOnMount` is disabled, it'll go to `handleSnapToIndex`.
- Usually, the first call to `handleSnapToIndex` will bail out due to [isLayoutReady](https://github.com/gorhom/react-native-bottom-sheet/blob/v5.2.7/src/components/bottomSheet/BottomSheet.tsx#L1085-L1088) condition.
[...]

PR #2594, opened 2026-01-10
Fix animatedPosition initialization logic

kliakhovskii-brex:
## Motivation

Tldr - fixing the issue described here:
https://github.com/gorhom/react-native-bottom-sheet/issues/2582

This initializes the `animatedPosition` with the right value if `animateOnMount` is disabled. 

This avoids two issues:
- Fixes issue with animation still playing if Layout is treated to be ready right away (e.g. no handle component and known size)
- Fixes first frame rendering in case when Layout is not ready (e.g. so the position is open right away rather than close -> open jump)


kliakhovskii-brex:
@gorhom wonder if i can have 👀  here?

kliakhovskii-brex:
> @gorhom wonder if i can have 👀 here?

still looking for ✅

Issue #2390, opened 2025-07-31
[Bug]:  Unable to resolve module @shopify/flash-list when using @gorhom/bottom-sheet


zaid0088:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS

### What happened?

I'm encountering an error when trying to use @gorhom/bottom-sheet. It fails to resolve the module @shopify/flash-list, even though it seems to be an optional dependency.

This was working fine until I updated React Native from version 0.72 to 0.73. After the upgrade, the app throws the following error at runtime.

my package.json 

`{
     "react-native": "0.73.11",
   "react-native-gesture-handler": "2.16.1",
    "react-native-reanimated": "3.16.1",
}`

### Reproduction steps

Install @gorhom/bottom-sheet

Use BottomSheetFlatList or other scrollable components

Upgrade React Native from 0.72.x to 0.73.x

Run the app

Encounter the error above

### Reproduction sample

https://snack.expo.dev/@gorhom/bottom-sheet---issue-reproduction-template

### Relevant log output

```shell
Error: Unable to resolve module @shopify/flash-list from /Users/app/node_modules/@gorhom/bottom-sheet/lib/commonjs/components/bottomSheetScrollable/BottomSheetFlashList.js: 
@shopify/flash-list could not be found within the project or in these directories:
  node_modules/@gorhom/bottom-sheet/node_modules
  node_modules
  ../../../../node_modules
  17 | // we try to import it using metro optional import
  18 | try {
> 19 |   FlashList = require('@shopify/flash-list');
     |                        ^
  20 | } catch (_) {}
```

aslihanturkdonmez:
Have you found any solution?

KJ-GM:
any solution for this?

Issue #1968, opened 2024-10-14
[Bug]: Error for not having flashlist even when not using the flashlist component


dannyhw:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS, Android, Web

### What happened?

When using version 5 without flashlist installed this error causes the app to crash

You need to install FlashList first,yarn install @shopify/flash-list, js engine: hermes

Since the files are all barrel files every file will run even if you don't import them so this would only be fixed by

removing the export from:
src/components/bottomSheetScrollable/index.ts and here src/index.ts

### Reproduction steps

- setup bottom sheet and don't install flashlist
- import scrollable (not flashlist)
- crashes

### Reproduction sample

https://snack.expo.dev/@dannyhw/bottom-sheet---issue-reproduction-template

### Relevant log output

```shell
You need to install FlashList first, `yarn install @shopify/flash-list`
  Evaluating @gorhom/bottom-sheet.js
  Evaluating App.js
  Loading App.js
```

FernandoAcTr:
Same issue here

gorhom:
this should be resolved with [v5.0.2](https://github.com/gorhom/react-native-bottom-sheet/releases/tag/v5.0.2). feel free to re-open

sleekLancelot:
> this should be resolved with [v5.0.2](https://github.com/gorhom/react-native-bottom-sheet/releases/tag/v5.0.2). feel free to re-open

Hi, I still face the same issue on v5.0.2, i need it for the dynamic sizing bit.
`You need to install FlashList first, yarn install @shopify/flash-list, js engine: hermes`

dannyhw:
@sleekLancelot I recommend double checking you have got the right version (check lockfile) and clearing the cache because it should only throw the error now if you use the flashlist scrollable.

sleekLancelot:
> @sleekLancelot I recommend double checking you have got the right version (check lockfile) and clearing the cache because it should only throw the error now if you use the flashlist scrollable.

Had to delete it, reinstall and rebuild again. For some reason it works now.

PR #2620, opened 2026-02-27
fix: inline FlashListProps to remove @shopify/flash-list type dependency

YevheniiKotyrlo:
## Problem

The published `BottomSheetFlashList.d.ts` unconditionally imports from `@shopify/flash-list`:

```typescript
import type { FlashListProps } from '@shopify/flash-list';
```

Since `@shopify/flash-list` is not declared in `dependencies`, `peerDependencies`, or `optionalDependencies`, consumers who don't install flash-list get:

```
error TS2307: Cannot find module '@shopify/flash-list' or its corresponding type declarations.
```

The source has `// @ts-ignore` above the import, which suppresses the error during this package's build — but TypeScript strips `@ts-ignore` comments from emitted `.d.ts` files by design ([TypeScript #38628](https://github.com/microsoft/TypeScript/issues/38628), confirmed "Working as Intended" by Ryan Cavanaugh). The downstream type error was never addressed.

Related: [#2390](https://github.com/gorhom/react-native-bottom-sheet/issues/2390), [#1968](https://github.com/gorhom/react-native-bottom-sheet/issues/1968)

## Fix

I replaced the external type import with a minimal local interface:

```typescript
import type { FlatListProps } from 'react-native';

interface FlashListProps<T> extends FlatListProps<T> {
    estimatedItemSize?: number;
}
```

This eliminates the external type dependency entirely. The runtime `try/catch require('@shopify/flash-list')` is unchanged — only types are affected.

The local interface covers all props that consumers of `BottomSheetFlashList` realistically use (`data`, `renderItem`, `keyExtractor`, list header/footer components, `onEndReached`, `refreshing`, `horizontal`, `numColumns`, `extraData`, plus `estimatedItemSize`). The few flash-list-specific props omitted (`getItemType`, `overrideItemLayout`, `drawDistance`) are niche optimization hints. For consumers who have `@shopify/flash-list` installed, Metro resolves the real types from the package directly.

This is the same pattern used by [ts-jest PR #3147](https://github.com/kulshekhar/ts-jest/pull/3147) to solve the identical problem with an optional `esbuild` dependency.

### Optional peer dependency — does not work

Adding `@shopify/flash-list` to `peerDependenciesMeta: { optional: true }` tells the package manager not to warn, but TypeScript does not consult `peerDependenciesMeta` during module resolution. The `.d.ts` still imports from `@shopify/flash-list`, and `tsc` still fails with TS2307.

### Separate entry point — correct but disproportionate

Moving `BottomSheetFlashList` to a subpath export (`@gorhom/bottom-sheet/flash-list`) so the main entry point never loads its `.d.ts` — the [Jotai pattern](https://dev.to/dai_shi/how-jotai-specifies-package-entry-points-2h7n). This works but is a breaking change for existing consumers and restructures exports for a deprecated component.

## Validation

| Check | Result |
|-------|--------|
| `yarn typescript` (tsc --skipLibCheck --noEmit) | Pass |
| `yarn lint` (biome) | Pre-existing issues only; this PR **reduces** warnings by 1 (removes a `@ts-ignore`) |
| `bob build` + `copy-dts` | Pass — generated `.d.ts` has no `@shopify/flash-list` import |
| Runtime JS output | Identical behavior — `try/catch require('@shopify/flash-list')` preserved |
| Consumer type-check (skipLibCheck: false) | Pass — no TS2307 |

### DIFF
--- a/src/components/bottomSheetScrollable/BottomSheetFlashList.tsx
+++ b/src/components/bottomSheetScrollable/BottomSheetFlashList.tsx
@@ -1,13 +1,20 @@
-// @ts-expect-error
-import type { FlashListProps } from '@shopify/flash-list';
 import React, { forwardRef, memo, type Ref, useMemo } from 'react';
-import type { ScrollViewProps } from 'react-native';
+import type { FlatListProps, ScrollViewProps } from 'react-native';
 import type { AnimatedProps } from 'react-native-reanimated';
 import BottomSheetScrollView from './BottomSheetScrollView';
 import type {
   BottomSheetScrollableProps,
   BottomSheetScrollViewMethods,
 } from './types';
+
+/**
+ * Minimal subset of FlashListProps needed for BottomSheetFlashList.
+ * Defined locally to avoid requiring @shopify/flash-list as a dependency,
+ * since the runtime import is optional (try/catch require).
+ */
+interface FlashListProps<T> extends FlatListProps<T> {
+  estimatedItemSize?: number;
+}
 
 let FlashList: {
   FlashList: React.FC;


YevheniiKotyrlo:
@gorhom — friendly bump. This fixes a real TypeScript breakage for consumers who don't install `@shopify/flash-list` — the published `.d.ts` files import it unconditionally even though it's optional, causing TS2307 errors with `skipLibCheck: false`.

All quality gates pass (tsc, biome lint, bob build). Happy to address any feedback.

gorhom:
thanks @YevheniiKotyrlo for submitting this PR 🙌

PR #2637, opened 2026-03-23
fix: StyleSheet absoluteFill instead of deprecated absoluteFillObject

huextrat:

## Motivation

### DIFF
--- a/example/src/screens/integrations/map/BlurredBackground.tsx
+++ b/example/src/screens/integrations/map/BlurredBackground.tsx
@@ -18,10 +18,10 @@ const BlurredBackground = () => {
 
 const styles = StyleSheet.create({
   blurView: {
-    ...StyleSheet.absoluteFillObject,
+    ...StyleSheet.absoluteFill,
   },
   container: {
-    ...StyleSheet.absoluteFillObject,
+    ...StyleSheet.absoluteFill,
     borderTopLeftRadius: 10,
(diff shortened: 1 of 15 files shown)


src/types.d.ts
```diff
@@ -30,7 +30,6 @@ import type {
   ANIMATION_STATUS,
   GESTURE_SOURCE,
   KEYBOARD_STATUS,
-  SCROLLABLE_STATUS,
```
gorhom, line 33: why removing this ?

package.json
```diff
@@ -66,16 +65,12 @@
...
   "peerDependencies": {
     "@types/react": "*",
-    "@types/react-native": "*",
     "react": "*",
     "react-native": "*",
     "react-native-gesture-handler": ">=2.16.1",
     "react-native-reanimated": ">=3.16.0 || >=4.0.0-"
   },
   "peerDependenciesMeta": {
-    "@types/react-native": {
```
gorhom, line 76: could you please revert this change

yarn.lock
gorhom, line 1: could you please revert this file too

gorhom (changes requested):
i would recommend keeping this PR only for replacing the deprecated API not formatting the entire code base

gorhom (approved):

gorhom:
thanks @huextrat

PR #2653, opened 2026-04-15
chore(deps): uninstall deprecated @types/react-native

PHILLIPS71:
## Motivation

`@types/react-native` is deprecated as types are now bundled within `react-native` itself, so it has been removed as a peer dependency to resolve the deprecation warnings

see: https://www.npmjs.com/package/@types/react-native


PHILLIPS71:
bump

PR #2626, opened 2026-03-02
fix: add typeof guard for getBoundingClientRect in useBoundingClientRect

magrinj:
PR #2561 fixed the `unstable_getBoundingClientRect` path (line 58) by adding a `typeof === 'function'` guard, but the fallback `getBoundingClientRect` path (line 71) was missed, it still only checks `!== null`.

In `react-test-renderer`, `nativeFabricUIManager` is set to `{}` by React Native's jest setup (`react-native/jest/setup.js` line 36), which makes `isFabricInstalled()` return `true`. This causes the `useBoundingClientRect` hook to execute the Fabric code path. Since `getBoundingClientRect` is `undefined` (not `null`) on test renderer refs, the `!== null` check passes and the call crashes:

```
TypeError: ref.current.getBoundingClientRect is not a function at useBoundingClientRect.ts:73:34
```

This affects any jest test that renders a `BottomSheet` component deeply enough to trigger the hook.

Fixes #2625
Reproduction: https://github.com/magrinj/bottom-sheet-test-bug

PR #2654, opened 2026-04-15
Fix bug crashed on iOS [new Architecture]

huydosgtech:

## Motivation


gorhom:
thanks for submitting this PR, but it is already addressed here #2626, it should be release shortly

Issue #2632, opened 2026-03-13
[Bug]: `BottomSheetModal` does not open after `dismissAll` `5.2.8`


rozhkovs:
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS

### What happened?

If we open several modals using `BottomSheetModal` and then call `dismissAll()`, the first modal may stop opening. This reproduces fairly consistently if done quickly, especially on iOS.

### Reproduction steps

1. Run the example from the repository.
2. Click the "Stack Modals" button.
3. Press "Present Modal A"
4. Press "Present Modal B"
5. Press "Dismiss All Modals"
6. Press "Present Modal A" again

You need to do it quickly enough for it to reproduce consistently.

For our application, this was critical; the temporary workaround was to switch to `stackBehavior={'replace'}`.

https://github.com/user-attachments/assets/d6cca838-9199-49e6-9af6-079a8fac10fb

### Reproduction sample

https://snack.expo.dev/@sergeyrozhkov/blank

### Relevant log output

```shell
LOG  [handlePresent] 
 LOG  [BottomSheetHandleContainer::handleBoundingClientRect] height:60
 LOG  [BottomSheetHandleContainer::handleContainerLayout] height:60
 LOG  [BottomSheet::useAnimatedReaction::OnSnapPointChange] result:568.5,379
 LOG  [BottomSheet::animateToPosition] currentPosition:874 nextPosition:568.5 source:1
 LOG  [BottomSheet::handleOnAnimate] toIndex:0 toPosition:568.5 fromIndex:-1 fromPosition:874
 LOG  [BottomSheet::useAnimatedReaction::OnSnapPointChange] result:568.5,379
 LOG  [BottomSheet::animateToPosition] currentPosition:874 nextPosition:568.5 source:1
 LOG  [handleMinimize] minimized:false
 LOG  [BottomSheet::handleClose] 
 LOG  [handlePresent] 
 LOG  [BottomSheet::animateToPosition] currentPosition:569.1839184857414 nextPosition:758 source:3
 LOG  [BottomSheet::handleOnAnimate] toIndex:-1 toPosition:758 fromIndex:-1 fromPosition:569.1839184857414
 LOG  [BottomSheetHandleContainer::handleBoundingClientRect] height:60
 LOG  [BottomSheetHandleContainer::handleContainerLayout] height:60
[...]

PR #2664, opened 2026-04-30
fix(modal): restore React mount state reset in unmount() after #2632

huextrat:
## Summary

Fixes a regression introduced in **v5.2.11** with [#2632](https://github.com/gorhom/react-native-bottom-sheet/issues/2632) (modal status refactor): after dismissing a `BottomSheetModal`, React state could keep `mount: true`, so a later `present()` (e.g. from `useFocusEffect` when navigating) would hit the fast path and call `snapToIndex(index)`—closed sheets appeared to reopen.

## Root cause

In **v5.2.10**, `unmount()` saved whether the portal was mounted **before** `resetVariables()`:

```ts
const _mounted = mounted.current; // mirrored `mount` each render
resetVariables();
// ...
if (_mounted) {
  setState(INITIAL_STATE);
}
```

In **v5.2.11**, the equivalent branch checked `statusRef` **after** `resetVariables()`, which always sets `statusRef` to `INITIAL`. The guard therefore never matched and `setState(INITIAL_STATE)` was skipped.

The condition also only considered `PRESENTED` and `ANIMATING`, while a completed dismiss ends as `DISMISSED`—so it would be wrong even if evaluated before reset.

## Fix

Mirror `mount` into `mountRef` on every render (same pattern as `mounted.current = mount` in v5.2.10). In `unmount()`, read `hadReactMount = mountRef.current` **before** `resetVariables()`, then call `setState(INITIAL_STATE)` when `hadReactMount` is true.

## Testing

- `yarn lint` (biome) on `./src` passes.
- `yarn typescript` still reports existing errors in other files on this branch (unchanged by this PR).

## Checklist

- [x] Minimal, targeted change
- [ ] Maintainer may want a repro example or regression test for modal + navigation

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/components/bottomSheetModal/BottomSheetModal.tsx
+++ b/src/components/bottomSheetModal/BottomSheetModal.tsx
@@ -64,6 +64,8 @@ function BottomSheetModalComponent<T = never>(
   //#region state
   const [{ mount, data }, setState] =
     useState<BottomSheetModalState<T>>(INITIAL_STATE);
+  const mountRef = useRef(mount);
+  mountRef.current = mount;
   //#endregion
 
   //#region hooks
@@ -109,6 +111,8 @@ function BottomSheetModalComponent<T = never>(
           method: unmount.name,
         });
       }
+      const hadReactMount = mountRef.current;
+
       // reset variables
       resetVariables();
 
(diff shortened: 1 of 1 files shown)


dorupopacaliman:
I can confirm 5.2.11 made my entire app unusable, this needs to be fixed asap

gorhom:
thanks @huextrat for submitting this PR, tested it and going to publish a new release shortly

PR #2665, opened 2026-04-30
fix: do not block mount animation when only keyboard is animating (#2661)

### LINKED ISSUE #2661: [Bug]: Keyboard overlapping bottom sheet with autofocus
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS, Android

### What happened?

I had a bug with version `5.2.6` on Android where the bottom sheet modal wouldn't dismiss. After investigating the issue, I found out it was resolved with version `5.2.10`. However, that version introduced a new bug where my bottom sheet content was being hidden by the keyboard.

I continued investigating and found out a similar bug by https://github.com/gorhom/react-native-bottom-sheet/issues/2658. However, I found out later that it was caused by the usage of the `autoFocus` prop on my `BottomSheetTextInput`. If I use that prop, it looks like the keyboard animation conflicts with the keyboard pushing up my content. If I delay the focus until the bottom sheet has the proper position or if I manually focus in my text input, the bug doesn't appear (as demonstrated in the Snack below).

While opening this issue, I saw the version `5.2.11` was released 3 minutes ago and I tested my reproduction snack and the issue is still present.

### Reproduction steps

- Having a `BottomSheetModal` that contains a BottomSheetTextInput with `autoFocus`
- Present the `BottomSheet`
- Notice the content not being pushed up by the keyboard and having the content hidden behind it
------------
For the Snack with the reproduction sample:
- I have 3 buttons which represent different scenarios
1. Bugged version with `autoFocus={true}`. We see that the keyboard raises and the content is being hidden behind it
2. Good (manual) version with `autoFocus={false}`. When we focus the `BottomSheetTextInput`, the keyboard raises properly and content is properly pushed above it
3. Patched workaround where we wait until the BottomSheet is fully raised (triggered by the `onChange`) until we manually focus the `BottomSheetTextInput`. The animation isn't smooth but it works.

I would expect the Bugged version to work out of the box as it did in version `5.2.6` and still having the bug resolution from https://github.com/gorhom/react-native-bottom-sheet/pull/2655

### Reproduction sample

https://snack.expo.dev/@glamontagne9/bottom-sheet---5-2-10

### Relevant log output

```shell

```

huextrat:
## Problem

Fixes https://github.com/gorhom/react-native-bottom-sheet/issues/2661

With **BottomSheetModal** + **BottomSheetTextInput** + **`autoFocus`**, the keyboard can open while `didAnimateOnMount` is still false. In that situation the sheet content can stay **behind** the keyboard. Focusing manually after the sheet is open (or after `onChange`) avoids the issue—a race between **mount** and **keyboard** animations.

## Root cause

PR https://github.com/gorhom/react-native-bottom-sheet/pull/2655 (`ad41447`, v5.2.10) added an early return in `evaluatePosition` for the mount phase:

```ts
if (!didAnimateOnMount.value) {
  if (animationStatus === ANIMATION_STATUS.RUNNING) {
    return;
  }
  // mount animation...
}
```

That correctly prevents the **mount** animation from overriding a **close** animation (e.g. `forceClose`), but it also treats **any** running animation the same way—including **keyboard** repositioning triggered by `autoFocus`. The mount path then never runs, and the sheet ends up misaligned with the keyboard.

## Fix

Only skip the mount animation when a **close** is in progress:

- `isForcedClosing` (force close / modal dismiss path), or
- `nextIndex === -1` (animating to the closed detent).

If the only `RUNNING` animation is keyboard-driven (`nextIndex` typically ≥ 0), the mount animation is **not** blocked.

## Testing

- `yarn biome check` on the touched file (via pre-commit / lint-staged).
- Manual: reproduce Snack from https://github.com/gorhom/react-native-bottom-sheet/issues/2661 — scenario with `autoFocus` should push content above the keyboard again.
- Regression: rapid `present` / `close` on `BottomSheetModal` (#2655) should still not freeze; close animations must still win over mount.

## Related

- Introduced by: #2655 (`ad41447`)
- Reported: #2661

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -967,10 +967,15 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
          */
         if (!didAnimateOnMount.value) {
           /**
-           * if there's a running animation (like force close), respect it and don't
-           * override with mount animation
+           * If a *close* animation is already running, do not override it with the mount
+           * animation (#2655). Keyboard-driven animations can also be RUNNING during this
+           * phase (e.g. TextInput autoFocus); those must not block mount or the sheet stays
+           * misaligned with the keyboard (#2661).
            */
-          if (animationStatus === ANIMATION_STATUS.RUNNING) {
+          if (
+            animationStatus === ANIMATION_STATUS.RUNNING &&
+            (isForcedClosing || nextIndex === -1)
+          ) {
             return;
           }
           /**


dorupopacaliman:
Confirming the patch works, this needs to be released.

gorhom (approved):
thanks @huextrat for submitting this PR

PR #2670, opened 2026-05-12
fix: make dismiss() idempotent on a never-presented modal (#2669)

### LINKED ISSUE #2669: [Bug]: dismiss() before first present() corrupts statusRef, blocking subsequent present()
### Version

v5 (5.2.14)

### Reanimated Version

v3 (the dropdown does not list v4 yet — actually tested on **4.2.1**, but the bug is pure JS state-machine logic so reanimated version is not load-bearing)

### Gesture Handler Version

v2 (2.30.1)

### Platforms

- iOS
- Android

### What happened?

`BottomSheetModal.dismiss()` is not safe to call before the first `present()`. If a consumer calls `dismiss()` on a `BottomSheetModal` instance whose status is still `MODAL_STATUS.INITIAL`, the modal's internal `statusRef` is set to `MODAL_STATUS.DISMISSING`. From that point on, **every subsequent `present()` call silently no-ops** — the Portal mounts but `handlePortalRender` bails out on the `DISMISSING` check, so the underlying sheet never renders. No warning, no error.

This breaks the canonical declarative-wrapper pattern that React users naturally reach for:

```tsx
useEffect(() => {
  if (visible) modalRef.current?.present();
  else modalRef.current?.dismiss();
}, [visible]);
```

On first mount `visible === false`, so `dismiss()` runs while gorhom is still in `INITIAL`. The subsequent `setVisible(true)` calls `present()`, but the sheet never appears.

### Reproduction steps

1. Render two identical `<BottomSheetModal>`s. Wire each to a button that calls `present()`.
2. On one of them, call `ref.current?.dismiss()` once from a `useEffect(() => { ... }, [])` on mount.
3. Tap the button for the *non*-dismissed sheet — it presents normally.
4. Tap the button for the dismissed-on-mount sheet — **nothing happens**.

### Reproduction sample

https://snack.expo.dev/@ycm.jason/bottomsheetmodal-v5-2-14-dismiss-before-first-present-is-broken

Two buttons side by side. The only difference between them is the `useEffect` that calls `dismiss()` on mount.

### Root cause

Verified by reading `src/components/bottomSheetModal/BottomSheetModal.tsx` on v5.2.14.

`handleDismiss` (L263) only early-exits for `[CLOSED, MINIMIZED]` and `(DISMISSING && currentIndexRef === -1)`. `INITIAL` is not in the list and falls through:

```ts
if (
  [MODAL_STATUS.CLOSED, MODAL_STATUS.MINIMIZED].includes(statusRef.current) ||
  (statusRef.current === MODAL_STATUS.DISMISSING && currentIndexRef.current === -1)
) {
  statusRef.current = MODAL_STATUS.DISMISSED;
  unmount();
  return;
}

statusRef.current = MODAL_STATUS.DISMISSING; // ← reached when status is INITIAL
willUnmountSheet(key);
bottomSheetRef.current?.forceClose(animationConfigs); // no-op: ref is null
```

`bottomSheetRef.current?.forceClose()` is a safe no-op because the underlying `BottomSheet` was never mounted — but `statusRef.current` is now stuck at `DISMISSING`. Nothing resets it. The `handleBottomSheetOnClose` callback (L496) that would normally transition `DISMISSING → DISMISSED` never fires, because there is no mounted `BottomSheet` to emit that callback.

Then in `handlePortalRender` (L411):

```ts
if ([MODAL_STATUS.DISMISSING].includes(statusRef.current)) {
  return;
}
render();
```

So when `present()` later mounts the Portal, `handlePortalRender` bails silently and the sheet never appears.

### Proposed fix

Add `MODAL_STATUS.INITIAL` to the early-exit list at L279. `dismiss()` on a never-presented modal should be idempotent (mirroring `present()`, which already handles the symmetric case):

```diff
  if (
-   [MODAL_STATUS.CLOSED, MODAL_STATUS.MINIMIZED].includes(statusRef.current) ||
+   [MODAL_STATUS.INITIAL, MODAL_STATUS.CLOSED, MODAL_STATUS.MINIMIZED].includes(statusRef.current) ||
    (statusRef.current === MODAL_STATUS.DISMISSING && currentIndexRef.current === -1)
  ) {
    statusRef.current = MODAL_STATUS.DISMISSED;
    unmount();
    return;
  }
```

### Workaround

Guard the dismiss call until the first present has happened:

```tsx
const hasPresentedRef = useRef(false);
useEffect(() => {
  if (visible) {
    modalRef.current?.present();
    hasPresentedRef.current = true;
  } else if (hasPresentedRef.current) {
    modalRef.current?.dismiss();
  }
}, [visible]);
```

### Relevant log output

_No errors or warnings logged in the failing case — that is part of what makes this a sharp footgun for downstream wrappers._

ycmjason:
Fixes #2669.

## Summary

`BottomSheetModal.handleDismiss` treats a never-presented modal (status `INITIAL`) as if it were mid-flight: it sets `statusRef` to `DISMISSING` and calls `forceClose` on a `bottomSheetRef` that is `null` because the inner `<Portal>` hasn't mounted. The forced-close never produces an `onClose` callback, so the status remains `DISMISSING` forever and every subsequent `present()` is silently swallowed by `handlePortalRender`'s DISMISSING-status check.

## Fix

Add `MODAL_STATUS.INITIAL` to the existing early-exit list in `handleDismiss`. For a never-presented modal, `dismiss()` becomes a clean no-op (status briefly `DISMISSED`, then `unmount()` resets it to `INITIAL` via `resetVariables()`), and `present()` works again.

## Reproduction (before this PR)

https://snack.expo.dev/@ycm.jason/bottomsheetmodal-v5-2-14-dismiss-before-first-present-is-broken

Or: render a `BottomSheetModal` and call `dismiss()` in a mount-time `useEffect`. The modal never opens on subsequent `present()` calls.

This shape is common in declarative wrappers that translate a `visible: boolean` prop into imperative `present()` / `dismiss()` calls — on first mount with `visible=false`, the cleanup-style call to `dismiss()` corrupts the state machine.

## Why the fix is safe

For an `INITIAL`-status modal entering the early-exit branch:
- `statusRef` is set to `DISMISSED` momentarily, then reset to `INITIAL` by `resetVariables()` inside `unmount()`.
- `unmountSheet(key)` is a no-op because the sheet was never in the provider's queue.
- `unmountPortal(key)` is a no-op.
- `setState(INITIAL_STATE)` is a no-op because state is already `INITIAL_STATE`.
- If `onDismiss` was provided, it fires once.

The net effect is "no state change, one onDismiss callback" — the correct semantic for dismissing a never-opened modal.

### DIFF
--- a/src/components/bottomSheetModal/BottomSheetModal.tsx
+++ b/src/components/bottomSheetModal/BottomSheetModal.tsx
@@ -273,8 +273,9 @@ function BottomSheetModalComponent<T = never>(
       }
 
       /**
-       * if the modal position is already in a closed position,
-       * then we unmount the node and early exit.
+       * if the modal has no inner sheet to animate (never presented,
+       * already closed, or minimized), then we unmount the node and
+       * early exit.
        */
       if (
         [
--- a/src/components/bottomSheetModal/BottomSheetModal.tsx
+++ b/src/components/bottomSheetModal/BottomSheetModal.tsx
@@ -277,9 +277,11 @@ function BottomSheetModalComponent<T = never>(
        * then we unmount the node and early exit.
        */
       if (
-        [MODAL_STATUS.CLOSED, MODAL_STATUS.MINIMIZED].includes(
-          statusRef.current
-        ) ||
+        [
+          MODAL_STATUS.INITIAL,
+          MODAL_STATUS.CLOSED,
+          MODAL_STATUS.MINIMIZED,
+        ].includes(statusRef.current) ||
         (statusRef.current === MODAL_STATUS.DISMISSING &&
           currentIndexRef.current === -1)
       ) {


ycmjason:
hi

ycmjason:
hi

ycmjason:
yo

olivier-bouillet:
Hey !

PR #2689, opened 2026-06-01
fix(layout): seed containerHeight on mount when raw->container reaction is missed (Reanimated 4)

### LINKED ISSUE #2688: BottomSheet mounts invisible on Android: containerHeight stuck at sentinel (Reanimated 4 reaction misses raw->container sync)
## Bug

On Android, a non-modal `BottomSheet` with an initial `index >= 0` intermittently mounts **invisible** — the component and its children are in the tree, but the sheet is positioned entirely off screen (at window height / closed) and never animates in. It happens most often on a cold app start when the dashboard/screen mounts while the JS thread is busy, so the container `onLayout` lands a second or two after mount. Under that contention it reproduces ~every time; in lighter conditions it's intermittent.

## Root cause

`useAnimatedLayout` keeps the measured container height in two fields of the layout state: `rawContainerHeight` (written directly by `BottomSheetHostingContainer`'s `onLayout`) and `containerHeight` (the effective value everything else reads). `containerHeight` is derived from `rawContainerHeight` by a `useAnimatedReaction`:

```ts
useAnimatedReaction(
  () => state.value.rawContainerHeight,
  (result, previous) => {
    if (result === previous) return;
    if (result === INITIAL_LAYOUT_VALUE) return;
    state.modify(_state => {
      'worklet';
      _state.containerHeight = modal ? result - verticalInset : result;
      return _state;
    });
  },
  [state, verticalInset, modal]
);
```

On **Reanimated 4** this reaction's handler can miss the initial `INITIAL_LAYOUT_VALUE (-999)` → measured-height transition (it does not fire for the first `layoutState.modify`). When that happens, `containerHeight` stays at the sentinel even though `rawContainerHeight` is correct.

Consequences cascade from there:
- `useAnimatedDetents` early-exits (`if (containerHeight === INITIAL_LAYOUT_VALUE) return {}`), so `detents` is `undefined`.
- `isLayoutCalculated` never becomes `true` (it requires normalized detents).
- The mount positioning reaction never runs `animateToPosition` / `setToPosition`, so `animatedPosition` stays at its initial `Dimensions.get('window').height` — i.e. fully off screen.

Notably, `useDerivedValue` consumers of the same layout state **do** recompute on the `modify`; only the `useAnimatedReaction` handler is skipped. That points at a Reanimated-4 reaction-vs-`modify` timing issue rather than app code.

## Captured evidence

Internal logging on a failing Android cold start (custom `print` added to `isLayoutCalculated`):

```
[BottomSheetHostingContainer::handleLayoutEvent] height:832
[isLayoutCalc] containerH=-999 rawContainerH=832 handleH=0 detents=none -> false
```

`rawContainerHeight = 832`, but `containerHeight` is still `-999`, detents never normalize, and `animateToPosition` is never called.

## Environment

- `@gorhom/bottom-sheet` 5.2.14 (current latest)
- `react-native-reanimated` 4.4.0
- `react-native-gesture-handler` 2.31.1
- React Native 0.81, Android (Samsung A54, Android 14)
- Non-modal `BottomSheet`, `enableDynamicSizing={false}`, `snapPoints={[<px>, '100%']}`, initial `index={0}`

## Reproduction

Mount a non-modal `BottomSheet` with `index={0}` as part of a screen that also kicks off heavy synchronous JS work at mount (so the container `onLayout` is delayed). On Android the sheet is frequently invisible until something forces a re-layout.

> Note: `animateOnMount={false}` is **not** a workaround — it makes it worse: it initializes `didAnimateOnMount = true`, and the layout-ready branch that calls `setToPosition` is gated on `!didAnimateOnMount`, so the sheet is never positioned at all (and `snapToIndex(0)` early-exits because `index === nextIndex`).

## Proposed fix

Seed a non-sentinel `containerHeight` in the hosting container's layout handler, in the same `modify` that writes `rawContainerHeight`, only while it is still the sentinel. The existing reaction keeps refining it (e.g. modal vertical inset) whenever it does run, so normal behavior is unchanged. PR attached.

christian-apollo:
## Summary

Fixes #2688.

On Android, a non-modal `BottomSheet` with an initial `index >= 0` can mount **invisible** — it renders into the tree but stays positioned off screen and never animates in. It's intermittent in general and reproduces reliably when the screen mounts under JS-thread contention (so the container `onLayout` lands a beat after mount).

### Root cause

`useAnimatedLayout` derives the effective `containerHeight` from `rawContainerHeight` via a `useAnimatedReaction`. On **Reanimated 4** that reaction can miss the initial `INITIAL_LAYOUT_VALUE` → measured-height transition (the handler doesn't run for the first `layoutState.modify`), so `containerHeight` is stuck at the sentinel even though `rawContainerHeight` is correct. From there `useAnimatedDetents` early-exits, `isLayoutCalculated` never turns `true`, and the mount positioning reaction never runs `animateToPosition`/`setToPosition` — the sheet is left at its initial `window.height` (off-screen) position.

`useDerivedValue` consumers of the same state recompute fine on the `modify`; only the `useAnimatedReaction` handler is skipped, which points at a Reanimated-4 reaction/`modify` timing issue.

Captured on a failing Android cold start:

```
[BottomSheetHostingContainer::handleLayoutEvent] height:832
isLayoutCalc containerH=-999 rawContainerH=832 handleH=0 detents=none -> false
```

### Change

Seed a non-sentinel `containerHeight` in `BottomSheetHostingContainer`'s layout handler, in the same `modify` that writes `rawContainerHeight`, **only while it is still the sentinel**. The existing reaction keeps refining the value (e.g. modal vertical inset) whenever it does run, so normal behavior is unchanged when the reaction fires — this only unblocks the case where it doesn't.

```ts
state.rawContainerHeight = height;
if (state.containerHeight === INITIAL_LAYOUT_VALUE) {
  state.containerHeight = height;
}
```

## Test plan

- Verified in a real RN 0.81 app (Reanimated 4.4.0, GH 2.31.1) on Android (Samsung A54): before the change the dashboard sheet was invisible on every cold launch; after, it positions to `index 0` on every launch (`animateToPositionCompleted … nextIndex:0`) and renders correctly.
- Behavior unchanged when the raw→container reaction fires normally (the seed is sentinel-guarded; the reaction still owns the final value, including the modal inset path).

## Notes

`animateOnMount={false}` is not a workaround for this — it independently prevents positioning (it sets `didAnimateOnMount = true`, and the layout-ready branch that calls `setToPosition` is gated on `!didAnimateOnMount`).

### DIFF
--- a/src/components/bottomSheetHostingContainer/BottomSheetHostingContainer.tsx
+++ b/src/components/bottomSheetHostingContainer/BottomSheetHostingContainer.tsx
@@ -8,6 +8,7 @@ import {
   View,
   type ViewStyle,
 } from 'react-native';
+import { INITIAL_LAYOUT_VALUE } from '../../constants';
 import { print } from '../../utilities';
 import { styles } from './styles';
 import type { BottomSheetHostingContainerProps } from './types';
@@ -61,6 +62,16 @@ function BottomSheetHostingContainerComponent({
         layoutState.modify(state => {
           'worklet';
           state.rawContainerHeight = height;
+          // `useAnimatedLayout` mirrors `rawContainerHeight` into `containerHeight`
+          // through a `useAnimatedReaction`. On Reanimated 4 that reaction can miss
+          // the initial `INITIAL_LAYOUT_VALUE` -> height transition (the handler does
+          // not fire for the first `modify`), leaving `containerHeight` at the
+          // sentinel so snap points never normalize and the sheet stays off screen.
+          // Seed a non-sentinel value here so layout can complete; the reaction still
+          // refines it (e.g. modal vertical inset) whenever it runs.
+          if (state.containerHeight === INITIAL_LAYOUT_VALUE) {
+            state.containerHeight = height;
+          }
           return state;
         });
       }


krisgerhard:
We're running this as a patch for last two weeks and it has solved our issues. Thank you!

bensalmon-daylight:
This PR shouldn't be closed, the issue still exists.

PR #2692, opened 2026-06-02
fix: preserve list callback types for bottom sheet scrollables

matsola-preply:
## Motivation

`BottomSheetFlatList<T>` and `BottomSheetSectionList<ItemT, SectionT>` currently wrap React Native list props with `AnimatedProps`, which can lose contextual typing for generic callback props.

As a result, even when users provide explicit list generics, TypeScript may treat callback parameters like `item` or `section` as implicitly `any` under `noImplicitAny`.

This change preserves the original React Native callback prop typings for both list components:

- `BottomSheetFlatList`: `keyExtractor`, `renderItem`
- `BottomSheetSectionList`: `keyExtractor`, `renderItem`, `renderSectionHeader`, `renderSectionFooter`

The rest of the props still use `AnimatedProps`, so existing animated prop support is preserved while list callback inference works as expected.


matsola-preply:
@gorhom do you have a chance to check it?

Issue #2639, opened 2026-03-25
[Bug]: Bottom sheet visible when closed on web at browser zoom levels below 100%


jspizziri:
### Version

v5.2.8

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

Web

### What happened?

When a `BottomSheet` with `index={-1}` (closed) is rendered on web, the sheet can become partially visible at the bottom of the screen when the browser zoom is below 100%.

### Cause

`INITIAL_POSITION` is set to `Dimensions.get('screen').height` (`screen.height`), which represents the display's physical resolution in CSS pixels and does not account for browser zoom. At zoom levels below 100%, the CSS pixel viewport (`window.innerHeight`) can exceed `screen.height`. The sheet is positioned at `translateY(INITIAL_POSITION)`, which is no longer off-screen.

The issue is more likely to manifest when the BottomSheet's container does not fill the full viewport (e.g., due to a navigation header or other layout elements), but it can occur in any layout where the viewport exceeds `screen.height`.

Additionally, when the sheet is closed and the container resizes (e.g., window resize), the closed position is not updated, leaving the sheet partially visible if the container grows beyond the current `translateY` value.

### Demo Video

You can see this in the attached screen recording using the example apps `FlatList` example with `index={-1}`. Upon navigating to the example the sheet should be closed and it should never be visible. This largely works at zoom=100% (including on window resize). However, zooms < 100% will cause the bottom sheet to become visible.

https://github.com/user-attachments/assets/7e1c9ee5-ada6-4c13-bfca-b7b59db9f23d

### Reproduction steps

1. In `example/src/screens/basic/BasicExamples.tsx`, change the BottomSheet `index` prop from `1` to `-1`:
    ```diff
    -          index={1}
    +          index={-1}
    ```
2. Open in a web browser and set zoom to 50%
  - NOTE: I've primarly reproduced this in Chrome and other Chromium browsers.
3. Make the height of the window "short"
4. Run the example app on web and navigate to any Basic example (e.g., "FlatList")
5. The bottom sheet handle/content is visible at the bottom of the screen despite index={-1}

### Reproduction sample

https://snack.expo.dev/@jspizziri-5stones/bottom-sheet---zoom-with-visible-bottom-sheet

### Relevant log output

```shell

```

gorhom:
Thanks @jspizziri for detailing the issue and submitting a potential fix , but i could not manage to repro the issue

[...]

PR #2667, opened 2026-05-05
feat(v6): introducing React Native Sheet

gorhom:
## 🚀 Introducing React Native Sheet (v6)

This PR introduces a complete rewrite of the library to leverage the latest React Native technologies. This version is built from the ground up to support the **New Architecture (Fabric/TurboModules)**, **Reanimated 4**, and **Gesture Handler v3**, providing a highly performant and flexible primitive for "Sheets" on all platforms.

### 📦 Rebranding: Bottom Sheet ⮕ Sheet
The library is being renamed to **React Native Sheet**. While "Bottom Sheet" was our origin, the new architecture allows the component to be anchored to any edge of the screen or behave as a free-floating modal. 

### 🛠 Key Changes & Features

*   **New Architecture Support:** Full compatibility with Fabric and TurboModules for synchronous layout calculations and improved performance.
*   **Reanimated & RNGH Integration:** Utilizes the latest worklet-based APIs for fluid, 60/120fps interactions.
*   **Layout-Driven Logic:** Rewritten internal logic that relies on React Native's layout engine rather than manual offset calculations, making it more robust.
*   **Native iPad & Landscape Support:** Improved responsiveness and adaptive layouts. Sheets now behave predictably when orientation changes or when used in Split View on iPad.
*   **Unstyled & Customizable:** We’ve moved toward a "Headless-ish" approach. Minimum default behaviors are enforced, giving developers full control over the UI, backdrop, and handle interactions.
*   **Flexible Positioning:** No longer constrained to the bottom; supports various presentation styles (Side sheets, Top sheets, or centered).

### ⚠️ Breaking Changes
*   **Package Name:** Transitioning from `@gorhom/bottom-sheet` to `@gorhom/sheet`.
*   **Default UI:** Many default styles have been stripped to allow for better customization.

### 📝 To-Do / WIP
- [ ] Draft a technical design document.
- [ ] Update documentation site with the new branding.
- [ ] Add migration guide for v5 users.
- [ ] Finalize E2E testing for Fabric.

PR #2697, opened 2026-06-05
fix: use screen height as initial position on native platforms

balintant:
## Motivation

On Android devices with a virtual navigation bar (common on Android 14 and older), `Dimensions.get('window').height` excludes the navigation bar height, making it shorter than the physical screen.

Using it as the initial off-screen position causes the sheet to appear partially visible on mount before any animation runs, with its backdrop blocking all interaction. `screen.height` is the safe upper bound on native; `window.height` is preserved on web where the distinction does not apply.

This issue was introduced in [v5.2.11](https://github.com/gorhom/react-native-bottom-sheet/releases/tag/v5.2.11) by the fix for #2639 ([17ee221](https://github.com/gorhom/react-native-bottom-sheet/commit/17ee22176123f94a17f2041a60ada3017f581463)).

**Note:** this issue is difficult to replicate on Android emulators, as newer Android versions no longer use a docked virtual navigation bar by default.

### DIFF
--- a/src/components/bottomSheet/BottomSheet.tsx
+++ b/src/components/bottomSheet/BottomSheet.tsx
@@ -215,7 +215,12 @@ const BottomSheetComponent = forwardRef<BottomSheet, BottomSheetProps>(
     const animatedCurrentIndex = useReactiveSharedValue(
       animateOnMount ? -1 : _providedIndex
     );
-    const animatedPosition = useSharedValue(Dimensions.get('window').height);
+    const animatedPosition = useSharedValue(
+      Platform.select({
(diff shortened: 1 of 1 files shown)


dmitrybirin:
@gorhom Hello! We've also bumped into this issue and patching it out exactly like @balintant. This is a valid fix, would be great to reopen

balintant:
@dmitrybirin I really don't think this (version of the) project is maintained anymore. We have switched over to react-native-action-sheet in the meantime and it works great (also has solution for sub-navigations, etc.).

v6: https://github.com/gorhom/react-native-bottom-sheet/pull/2667

Issue #701, opened 2021-10-17
[v4] Entire view being blocked by BottomSheetBackdrop


Prabindas001:
# Bug

The first time it works as expected, but once I move to some other screen and come back, all touch events are blocked by the backdrop overlay. Though, the backdrop component is not visible at that time. 

The same app is tested on Android and iOS, but the issue is only present on the Android device.

## Environment info

| Library                         | Version |
| ------------------------------- | ------- |
| @gorhom/bottom-sheet            | 4   |
| react-native                    | expo: sdk-42.0.0   |
| react-native-reanimated         | 2.2.0   |
| react-native-gesture-handler    | 1.10.2   |

### Device Details:
**OS:** Android 11 (MIUI Global 12.5.8 Stable)
**Model:** Redmi Note 10 Pro (M2101K6P)

## Steps To Reproduce

1. Open the bottom sheet and close it
2. Change to some other using stack navigator
3. Come back to this page, touch events should be blocked.

## Reproducible sample code

Its all from example code, still I'm adding the code here:

```
  const bottomSheetRef = useRef();
  const snapPoints = useMemo(() => ["80%"], []);

  const renderBackdrop = useCallback(
    (props) => (
      <BottomSheetBackdrop
        {...props}
        disappearsOnIndex={-1}
        appearsOnIndex={0}
        enableTouchThrough={true}
      />
    ),
    []
  );

  return (
  <KeyboardAvoidingView
      style={styles.groceriesParentView}
      behavior={Platform.OS === "ios" ? "padding" : null}>
      <SafeAreaView style={styles.container}>
        
        <BottomSheet
          ref={bottomSheetRef}
          index={-1}
          snapPoints={snapPoints}
          onChange={handleSheetChanges}
          enablePanDownToClose={true}
          backdropComponent={renderBackdrop}>
             <View>...<View/>
        </BottomSheet>

      </SafeAreaView>
    </KeyboardAvoidingView>
)
```

nyoung697:
Are you seeing this log show up on your device by any chance?

> Could not locate shadow view with tag #1065, this is probably caused by a temporary inconsistency between native views and shadow views.

I am experiencing a similar issue to yours I think.  It is happening on iOS though.  It seems to be working okay on Android.

It is inconsistent with when it happens though.  However, when it does occur, I get that message printed out in the device logs.

[...]

PR #2706, opened 2026-06-29
Fixes an Android bug where a closed BottomSheet with a backdrop can block all touch events on the underlying screen after a cold start.

### LINKED ISSUE #2680: [Bug]: Invisible closed backdrop can block touches after cold start on some Android devices
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

Android

### What happened?

After upgrading `@gorhom/bottom-sheet` from `5.2.8` to `5.2.14`, we started seeing an Android issue where a screen behind a closed `BottomSheet` stopped receiving touch events after a cold app start.

Important context: I personally did not have a physical device where the issue reproduced, so debugging was based on reports from colleagues and version bisecting. However, on affected devices the issue reproduced consistently.

Affected devices:

- Samsung, Android 14
- Xiaomi, Android 11

It did not reproduce for me on emulator or on a newer Samsung Android device.

The affected screen renders a normal `ScrollView` and also mounts a `BottomSheet` in the closed state with `index={-1}`. On cold start, scrolling and button presses on the screen did not work. Visually the screen was rendered correctly, but touches seemed to be swallowed. If the user switched to another app and then returned to our app, scrolling and button presses started working again without navigating away.

This suggests a possible race condition around `BottomSheetBackdrop` touchability. Relevant code path:

```ts
const [pointerEvents, setPointerEvents] = useState<ViewProps['pointerEvents']>(
  enableTouchThrough ? 'none' : 'auto'
);
```

The backdrop starts with `pointerEvents='auto'`, then `useAnimatedReaction` should switch it to `none` when the sheet is closed. But `handleContainerTouchability` only updates state when `isMounted.current === true`. If the reaction fires before the React `useEffect` sets `isMounted.current = true`, the update can be skipped. Later, since `animatedIndex` is still `-1`, there may be no new reaction until the app resumes or layout/dimensions change.

The issue appeared after moving from `5.2.8` to `5.2.14`. I suspect changes around initial position/layout timing, especially the move to `Dimensions.get('window').height`, made this timing issue visible on some Android devices.

### Reproduction steps

1. Install `@gorhom/bottom-sheet@5.2.14`.
2. Create a screen with scrollable content and a closed bottom sheet mounted immediately.
3. Use a backdrop with `appearsOnIndex={0}` and `disappearsOnIndex={-1}`.
4. Cold start the app on an affected physical Android device.
5. Try to scroll or press buttons on the screen behind the closed sheet.
6. Observe that touches are blocked consistently on affected devices.
7. Switch to another app and return to the app.
8. Observe that scrolling and button presses start working again.

Minimal example:

```tsx
import React, {useCallback, useRef} from 'react';
import {Button, ScrollView, Text} from 'react-native';
import {GestureHandlerRootView} from 'react-native-gesture-handler';
import BottomSheet, {BottomSheetBackdrop, BottomSheetScrollView} from '@gorhom/bottom-sheet';

export default function App() {
  const sheetRef = useRef<BottomSheet>(null);

    props => (
      <BottomSheetBackdrop
        {...props}
        appearsOnIndex={0}
        disappearsOnIndex={-1}
      />
    ),
    [],
  );

  return (
    <GestureHandlerRootView style={{flex: 1}}>
      <ScrollView contentContainerStyle={{padding: 24}}>
        {Array.from({length: 40}).map((_, index) => (
          <Text key={index} style={{paddingVertical: 12}}>
            Scroll row {index + 1}
          </Text>
        ))}

        <Button title="Open sheet" onPress={() => sheetRef.current?.snapToIndex(0)} />
      </ScrollView>

      <BottomSheet
        ref={sheetRef}
        index={-1}
        enableDynamicSizing
        enablePanDownToClose
        backdropComponent={renderBackdrop}>
        <BottomSheetScrollView contentContainerStyle={{padding: 24}}>
          <Text>Bottom sheet content</Text>
        </BottomSheetScrollView>
      </BottomSheet>
    </GestureHandlerRootView>
  );
}
```

Expected: when `index={-1}`, the closed sheet/backdrop should not block touches behind it.

Actual: on affected Android devices after cold start, the screen behind the closed sheet does not receive scroll/touch events until the app goes to background and returns.

### Reproduction sample

https://snack.expo.dev/@gorhom/bottom-sheet---issue-reproduction-template

### Relevant log output

```shell

```

kanav22:
Fixes an Android bug where a closed BottomSheet with a backdrop can block all touch events on the underlying screen after a cold start.

Closes #2680

Problem
When a screen mounts a closed bottom sheet (index={-1}) with a BottomSheetBackdrop, some Android devices stop receiving scroll and button presses on the content behind the sheet. The UI renders correctly, but touches appear to be swallowed. Backgrounding the app and returning restores interaction without navigating away.

Reported on devices including Samsung (Android 14) and Xiaomi (Android 11).

Root cause
Two independent issues in BottomSheetBackdrop:

1. Mount race condition
pointerEvents was always initialized to 'auto' when enableTouchThrough is false. useAnimatedReaction can fire before the mount useEffect sets isMounted.current = true. Because handleContainerTouchability only updates state when mounted, that first update is silently skipped and the invisible backdrop keeps intercepting touches.

2. Floating-point animatedIndex
On some devices, animatedIndex settles at -0.999… instead of exactly -1. With disappearsOnIndex={-1}, the check animatedIndex.value <= disappearsOnIndex evaluates to false, so pointerEvents never switches to 'none'.

Solution
Derive the initial pointerEvents value from the current animatedIndex instead of defaulting to 'auto'
Sync pointerEvents on mount to recover from the race where the animated reaction fires before the component is mounted
Use an epsilon-based closed-state check (index <= disappearsOnIndex + 0.01) so near-closed values like -0.9999 are treated as closed without affecting partially open positions during animations
How to reproduce (before fix)
Use @gorhom/bottom-sheet@5.2.14
Render a screen with scrollable content and a closed bottom sheet (index={-1})
Add a backdrop with appearsOnIndex={0} and disappearsOnIndex={-1}
Cold start on an affected Android device
Try scrolling or pressing buttons behind the sheet
Observe that touches are blocked until the app is backgrounded and resumed
<BottomSheet
  ref={sheetRef}
  index={-1}
  enableDynamicSizing
  enablePanDownToClose
  backdropComponent={renderBackdrop}
>
  ...
</BottomSheet>
Test plan

yarn typescript
passes

biome check
passes on the changed file
Cold-start repro on physical Android device (Samsung / Xiaomi class devices)
Verify backdrop still captures presses when the sheet is open
Verify
enableTouchThrough
behavior is unchanged
Verify opening/closing the sheet still works as expected
Risks / notes
The epsilon value (0.01) is intentionally small so mid-animation indices (e.g. -0.3) are not incorrectly treated as closed
This is a targeted change to BottomSheetBackdrop.tsx only; no API changes
Related community workaround was setting disappearsOnIndex={-0.5} — this fix addresses the underlying comparison issue without requiring consumers to change their config

### DIFF
--- a/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
+++ b/src/components/bottomSheetBackdrop/BottomSheetBackdrop.tsx
@@ -30,6 +30,13 @@ import {
 import { styles } from './styles';
 import type { BottomSheetDefaultBackdropProps } from './types';
 
+const BACKDROP_INDEX_EPSILON = 0.01;
+
+const shouldDisableBackdropTouchability = (
+  index: number,
+  disappearsOnIndex: number
+) => index <= disappearsOnIndex + BACKDROP_INDEX_EPSILON;
+
 const BottomSheetBackdropComponent = ({
   animatedIndex,
   opacity: _providedOpacity,
@@ -62,7 +69,18 @@ const BottomSheetBackdropComponent = ({
   //#region variables
   const [pointerEvents, setPointerEvents] = useState<
     ViewProps['pointerEvents']
-  >(enableTouchThrough ? 'none' : 'auto');
+  >(() => {
+    if (enableTouchThrough) {
+      return 'none';
+    }
+
+    return shouldDisableBackdropTouchability(
+      animatedIndex.value,
+      disappearsOnIndex
+    )
+      ? 'none'
+      : 'auto';
+  });
   //#endregion
 
   //#region callbacks
@@ -120,7 +138,8 @@ const BottomSheetBackdropComponent = ({
 
   //#region effects
   useAnimatedReaction(
-    () => animatedIndex.value <= disappearsOnIndex,
+    () =>
+      shouldDisableBackdropTouchability(animatedIndex.value, disappearsOnIndex),
     (shouldDisableTouchability, previous) => {
       if (shouldDisableTouchability === previous) {
         return;
@@ -134,10 +153,20 @@ const BottomSheetBackdropComponent = ({
   // [link](https://github.com/gorhom/react-native-bottom-sheet/issues/1376)
   useEffect(() => {
     isMounted.current = true;
+
+    // Sync pointer events on mount in case the animated reaction fired before
+    // the component was mounted (https://github.com/gorhom/react-native-bottom-sheet/issues/2680)
+    const shouldDisableTouchability = shouldDisableBackdropTouchability(
+      animatedIndex.value,
+      disappearsOnIndex
+    );
+
+    setPointerEvents(shouldDisableTouchability ? 'none' : 'auto');
+
     return () => {
       isMounted.current = false;
     };
-  }, []);
+  }, [animatedIndex, disappearsOnIndex]);
   //#endregion
 
   const AnimatedView = (


jesselopespereira:
Any update when this PR will be merged?

carterbancroft:
Yes, also wondering when this will merge. This fixes the issue for me on Android 15 and in my case the issue arises not just on cold starts but consistently (even when backgrounding the app) and cannot be recovered from.

kasunmendis7:
I ran into this exact issue in my own project — the closed BottomSheet with a backdrop was blocking all touch events after a cold start.  
I had to patch it locally in my codebase to get around it, so I hope this issue taken into investigation and fixed.

kanav22:
@gorhom following up on this -there is still demand to merge it.

This PR is still mergeable against master (one file, no API change). Linked issue #2680 was closed as stale, not as fixed, and the bug is still present in 5.2.14. Recent comments here confirm it on Android 11–15 (Samsung, Xiaomi, Pixel); in at least one report it is consistent, not only on cold start, and people are patching locally to ship.

Related #2342 ([Math.floor](https://github.com/gorhom/react-native-bottom-sheet/pull/2706#) on animatedIndex) covers the same floating-point miss, but rounding the index too aggressively lets taps go through the backdrop while the sheet is still opening. The epsilon check here (index <= disappearsOnIndex + 0.01) treats near-closed values like -0.999… as closed without changing mid-animation behavior.

Happy to adjust if you want a different threshold or extra coverage. Otherwise this looks ready to land.

kartikraj:
spent 2 days debugging as the issue occured only for Redmi note 7 pro , Android 10  . Thanks for the PR @kanav22 
 
 Confirming this PR addresses a real bug, and flagging one defect that currently
makes it unusable on the New Architecture.

Environment

- react-native **0.86.2**, New Architecture (Fabric + bridgeless)
- react-native-reanimated  - 4.5.3 , react-native-worklets  0.11.3
- @gorhom/bottom-sheet 5.2.14
- Android only (reproduced on a Redmi Note 7 Pro, API 29); iOS unaffected

Blocker: `shouldDisableBackdropTouchability` needs a `'worklet'` directive

The helper is called from inside the `useAnimatedReaction` worklet, but it's a
plain module-scope function, so it stays a JS-thread ("remote") function. Every
screen that mounts a backdrop then throws :
[worklets] tried to synchronously call a remote function, called anonymous on the UI runtime

Fix:

```
```diff
 const shouldDisableBackdropTouchability = (
   index: number,
   disappearsOnIndex: number
-) => index <= disappearsOnIndex + BACKDROP_INDEX_EPSILON;
+) => {
+  'worklet';
+  return index <= disappearsOnIndex + BACKDROP_INDEX_EPSILON;
+};
```

Worklet-ised functions stay callable from the JS thread, so the useState
initialiser and the new mount effect keep working unchanged. With the directive
added, the plugin emits it as its own worklet and the reaction's call becomes a
UI-runtime call:

With the 'worklet' directive added, this PR is the only approach we found that
fixes it without giving up tap-to-close (a static pointerEvents="none" wrapper
works but kills the backdrop press behaviour, per @Rc85's objection in #701).

kanav22:
Thanks @kartikraj -`shouldDisableBackdropTouchability` is now a worklet, so the `useAnimatedReaction` call stays on the UI runtime and no longer throws on the New Architecture.
@gorhom this is still one file with no API change, and ready for review.

PR #2709, opened 2026-06-30
feat(accessibility): forward importantForAccessibility prop

vicprz:
## Motivation

I have been focusing heavily on improving the accessibility of my application and encountered a few blocking issues on Android with the current implementation.

### 1. Focusable BottomSheetContent
By default, the entire `BottomSheetContent` was being treated as a single focusable element. This prevented screen readers from navigating inside the container to interact with individual components. While passing `accessible={false}` to `BottomSheetModal` resolved the issue on iOS, Android requires `importantForAccessibility="no"`. This prop was not previously being forwarded by the library, so this PR adds that support.

### 2. Focusable Backdrop
I needed to prevent the backdrop from being focusable by screen readers since I have an explicit "Close" button inside the bottom sheet. Similar to the issue above, `accessible={false}` worked perfectly for iOS, but Android required `importantForAccessibility="no"`. This PR brings that property to the backdrop component.

### 3. Consistency across components
For the sake of consistency, I also added `importantForAccessibility` prop forwarding to `BottomSheetHandle`, as it was the last remaining place where `NullableAccessibilityProps` were being utilized without it.


vicprz:
Still relevant and ready for review whenever someone has a moment 🙂

vicprz:
Still relevant 🙂

PR #2725, opened 2026-07-28
fix(modal): stop present() silently no-opping forever (two separate wedges)

### LINKED ISSUE #2723: [Bug]: dismiss() on a not-presented BottomSheetModal permanently wedges status at DISMISSING — later present() calls silently no-op
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

iOS, Android

### What happened?

Calling `dismiss()` on a `BottomSheetModal` that is **not currently presented** — either never presented, or already self-dismissed by the user (backdrop tap / pan-down) — leaves the modal permanently broken: every subsequent `present()` mounts the Portal but renders nothing, with no error or warning. Only unmounting and remounting the modal component recovers.

This is easy to hit in real apps: any parent that mirrors "sheet open" into its own state and calls `dismiss()` from an effect/cleanup will eventually call it after the user already dismissed the sheet themselves (the modal's `onDismiss` → state update → effect runs → second `dismiss()`).

**Root cause (traced in 5.2.14 source, `src/components/bottomSheetModal/BottomSheetModal.tsx`):**

1. After a user self-dismissal, `handleBottomSheetOnClose` → `unmount()` resets `statusRef.current` to `INITIAL`, sets `mount: false` (so the inner `bottomSheetRef.current` becomes `null`), and fires `onDismiss` last. A never-presented modal is also at `INITIAL`.
2. A `dismiss()` call now falls through `handleDismiss`'s early-exit guard (~lines 279–289): the guard checks `CLOSED`, `MINIMIZED`, and `DISMISSING && currentIndexRef === -1`, but **not `INITIAL`**. So it sets `statusRef.current = MODAL_STATUS.DISMISSING` (~line 291) and calls `bottomSheetRef.current?.forceClose(...)` — a silent no-op on the null ref. Nothing is animating, so no `onClose` will ever fire, and nothing ever transitions the status out of `DISMISSING`.
3. Every later `present()` sets `mount: true`, but the Portal's `handlePortalRender` (~lines 399–415) returns early while `statusRef.current` is `DISMISSING` — `render()` is never invoked, so the sheet content is never injected into the host and no `onChange` fires that could repair the status. The provider queue also retains the key, so retried `present()` calls early-exit in `mountSheet`.

**Suggested fix** — either is a one-line guard: `handleDismiss` early-returns when `statusRef.current === MODAL_STATUS.INITIAL` (nothing to dismiss), and/or `handlePresent` clears a stale `DISMISSING` status before mounting so a wedged modal self-heals.

**Workaround:** track presented-ness in the parent and reset that flag inside the modal's `onDismiss` (which fires for *any* actual dismissal), so `dismiss()` is only ever called while the modal is genuinely presented.

(Environment note: observed on a physical Android device; the mechanism is pure JS/TS so it should affect all platforms. Reproduced with reanimated 4.5.0 / gesture-handler 2.32.0 / RN 0.86; the Reanimated dropdown caps at v3. Possibly related, closed unresolved: #2140, #1647, #343.)

### Reproduction steps

- Repro A (smallest): press `dismiss()` FIRST (modal never presented), then press `present()` → the sheet never renders, forever.
- Repro B (real-world path): press `present()` → tap the dark backdrop to close the sheet → press `dismiss()` (simulates a parent effect reacting to the `onDismiss` state update) → press `present()` → the sheet never renders again.

### Reproduction sample

https://snack.expo.dev/RG1RX3LXwGFUg8r7VEBwQ

### Relevant log output

```shell
(no output — the failure is silent: present() no-ops with no warning or error)
```

ronakbothraa:
## Summary

Two independent defects in `BottomSheetModal` that both end in the same silent failure: **`present()` mounts the portal but the sheet never appears, permanently, with no error or warning.** Only unmounting and remounting the component recovers.

They have different triggers and different code paths, but they live a few lines apart in the same lifecycle machinery, so they're fixed together here. Each is a separate commit and can be split if you'd prefer.

| | Trigger | Wedged status | Commit |
|---|---|---|---|
| **1** | `dismiss()` while the modal is not presented | `DISMISSING` | `49af5f6` |
| **2** | Orphaned portal entry survives teardown | `INITIAL` | `c9995c3` |

---

## 1. `dismiss()` on a non-presented modal — fixes #2723

Full credit to @Quraian for the report and the diagnosis; this implements the first of the two fixes suggested there.

`handleDismiss`'s early-exit guard checks `CLOSED`, `MINIMIZED`, and `DISMISSING && currentIndexRef === -1` — but not `INITIAL`. Both a never-presented modal *and* one that has already been torn down sit at `INITIAL` (`unmount()` resets the status and sets `mount: false`, so `bottomSheetRef.current` becomes `null`).

So a `dismiss()` in that state falls through, sets `statusRef.current = DISMISSING`, and calls `forceClose()` on a null ref — a silent no-op. Nothing is animating, so `onClose` never fires, so nothing transitions the status out of `DISMISSING`. `handlePortalRender` then suppresses every subsequent render.

This is easy to hit from any parent that mirrors "sheet open" into its own state: the modal's `onDismiss` → state update → effect → a second `dismiss()` after the user already dismissed it themselves.

**Fix:** there is nothing to dismiss at `INITIAL`, so return early.

## 2. Orphaned portal entry keeps the sheet mounted but unopenable

This one is separate from #2723 and, as far as I can tell, unreported. It survives the usual userland workaround for #2723 (tracking presented-ness in the parent), because here `dismiss()` is only ever called while genuinely presented.

`Portal`'s `handleOnUpdate` effect is keyed on `children`, whose *element identity* changes on every parent render. A trailing update can therefore land in the window between `unmount()`'s `removePortal` and the `Portal` component actually unmounting — **re-adding the entry that was just removed.**

`handlePortalOnUnmount` then early-returns at `INITIAL` without calling the `removePortal` callback that `Portal` hands it (the prop is typed `handleOnUnmount?: (unmount: () => void) => void` — it's deliberately provided for exactly this). So the re-added entry is orphaned permanently.

Consequences:

- The host keeps rendering the stale node, so the inner `BottomSheet` is **never unmounted** — every closed sheet leaks a live subtree.
- The next `present()` cannot open it either. `handlePresent` reads a captured `mount` of `false` and so skips the `snapToIndex` branch, while the still-mounted sheet already has `didAnimateOnMount === true` and therefore runs no mount animation. **Nothing ever asks the sheet to open.**

**Fix:** call the provided `removePortal` in the `INITIAL` branch. `removePortal` is idempotent (`findIndex → -1 → no-op`), so this does nothing when there is no orphan.

### Trace from a real app

Captured on device, `5.2.14`, New Architecture (Fabric), React 19, reanimated 4. Sheet had closed cleanly ~600ms earlier; note the status is `0 (INITIAL)` throughout — never `DISMISSING`, which is what distinguishes this from #2723:

```
BottomSheetModal::unmount
BottomSheetModal::resetVariables          ← statusRef := INITIAL
BottomSheetModal::handlePortalRender status:0
PORTAL action=REMOVE  before=[modal-110] after=[]        ← unmount() removed it
PORTAL action=ADD     before=[]          after=[modal-110]  ← re-added 1ms later
PORTAL host RENDER count=1 names=[modal-110]             ← orphan being rendered
BottomSheetModal::handlePortalOnUnmount status:0         ← early-returns, orphan kept

--- user taps to reopen ---
BottomSheetModal::handlePresent currentIndexRef:-1 status:0
BottomSheetModal::handlePortalRender status:0
   ...no inner mount, no layout, no onSnapPointChange, no animation. Sheet never appears.
```

## Verification

Honest breakdown of what is and isn't verified:

- **Defect 2 is verified end-to-end.** Root-caused from the device trace above (instrumented `evaluatePosition`, the portal reducer, and the host render), and confirmed fixed on device after the change. In our app it additionally turns **3 pre-existing red tests green** — `renders children if isVisible is true`, `should render scroll view if scrollable is true`, `should not render scroll view if scrollable is false` — with **zero** newly-failing tests (controlled A/B, stable across repeated runs).
- **Defect 1 is a code-level fix**, matching the mechanism traced in #2723 and the reporter's own suggestion. I have **not** run the Expo Snack repro myself, so please treat that as reviewed-not-executed.
- `yarn typescript` clean, `yarn lint` clean on the changed file (the one remaining repo warning is pre-existing in `src/hooks/useBoundingClientRect.ts`, identical on `master`), `yarn build` succeeds.
- **No tests added.** `CONTRIBUTING.md` mentions `yarn test`, but there's no `test` script or jest setup in the repo, so adding tests would mean introducing the whole harness. Very happy to do that in a separate PR if you'd like it.

## Notes for reviewers

- Both changes are **purely additive to branches that previously did nothing** (a bare `return`, and a missing guard). No existing code path changes behaviour.
- `statusRef` is `INITIAL` in exactly two situations: the initial value, and after `resetVariables()` — whose only caller is `unmount()`. So the new `handlePortalOnUnmount` behaviour can only ever fire on an already-torn-down sheet.
- Paths I checked as unaffected by change 2 (status is not `INITIAL` at cleanup time): backdrop tap, swipe-to-dismiss, unmounting the screen with the sheet open, `enableDismissOnClose={false}`, and rapid `dismiss()`→`present()` mid-close (the Portal never unmounts there).
- #2723 also suggests a second, complementary guard: having `handlePresent` clear a stale `DISMISSING` status so an already-wedged modal self-heals. I left it out to keep this focused, but it would add defence-in-depth for any residual path that can still strand the status (e.g. `handlePortalOnUnmount`'s non-`INITIAL` branch calling `close()` on a null ref). Happy to add it if you want.

### DIFF
--- a/src/components/bottomSheetModal/BottomSheetModal.tsx
+++ b/src/components/bottomSheetModal/BottomSheetModal.tsx
@@ -379,7 +379,7 @@ function BottomSheetModalComponent<T = never>(
 
   //#region callbacks
   const handlePortalOnUnmount = useCallback(
-    function handlePortalOnUnmount() {
+    function handlePortalOnUnmount(removePortalFromHost?: () => void) {
       if (__DEV__) {
         print({
           component: 'BottomSheetModal',
@@ -391,6 +391,27 @@ function BottomSheetModalComponent<T = never>(
       }
 
       if (statusRef.current === MODAL_STATUS.INITIAL) {
+        /**
+         * The sheet is already fully torn down — `unmount()` removed this portal
+         * from the host and reset the status.
+         *
+         * However `Portal`'s `handleOnUpdate` effect is keyed on `children`, whose
+         * element identity changes on every parent render. A trailing update can
+         * therefore land in the window between `unmount()`'s `removePortal` and
+         * this component actually unmounting, RE-ADDING the entry we just removed.
+         *
+         * Left behind, that orphan makes the host keep rendering the stale node,
+         * so the inner `BottomSheet` is never unmounted. The next `present()` then
+         * cannot open it: `handlePresent` reads a captured `mount` of `false`, so
+         * it skips the `snapToIndex` branch, while the still-mounted sheet already
+         * has `didAnimateOnMount === true` and so runs no mount animation either.
+         * The sheet silently never opens again, and every closed sheet leaks a
+         * live subtree.
+         *
+         * This is the last chance to drop that orphan. `removePortal` is
+         * idempotent, so this is a no-op when there is nothing stale to remove.
+         */
+        removePortalFromHost?.();
         return;
       }
 
--- a/src/components/bottomSheetModal/BottomSheetModal.tsx
+++ b/src/components/bottomSheetModal/BottomSheetModal.tsx
@@ -273,6 +273,23 @@ function BottomSheetModalComponent<T = never>(
       }
 
       /**
+       * if the modal was never presented, or has already been fully torn down
+       * (`unmount()` resets the status back to `INITIAL`), then there is nothing
+       * to dismiss and no inner sheet to close.
+       *
+       * Falling through would set the status to `DISMISSING` and call
+       * `forceClose()` on a `bottomSheetRef` that is still `null` — a silent
+       * no-op. Since nothing is animating, `onClose` never fires, so nothing ever
+       * transitions the status out of `DISMISSING`, and `handlePortalRender`
+       * suppresses every subsequent render. The modal is then permanently wedged:
+       * later `present()` calls mount the portal but render nothing, with no
+       * error or warning.
+       */
+      if (statusRef.current === MODAL_STATUS.INITIAL) {
+        return;
+      }
+
+      /**
        * if the modal position is already in a closed position,
        * then we unmount the node and early exit.
        */


JorgeArmandoMT:
Works for me,  thanks @ronakbothraa !

PR #2759, opened 2026-08-30
fix: omit native scroll handler dependencies

### LINKED ISSUE #2758: [Bug]: BottomSheetScrollView triggers Reanimated 4.6.0 native dependency warning
### Version

v5

### Reanimated Version

v3

### Gesture Handler Version

v2

### Platforms

Android

### What happened?

After upgrading `react-native-reanimated` to 4.6.0, mounting `BottomSheetScrollView` produces the following warning in native development builds:

```text
[Reanimated] dependencies should only be used in web implementation.
```

The bottom sheet remains functional, but the warning is emitted whenever the scrollable component mounts.

## Environment

| Library | Version |
| --- | --- |
| `@gorhom/bottom-sheet` | 5.2.14 |
| `react-native` | 0.86.3 |
| `react` | 19.2.3 |
| `react-native-reanimated` | 4.6.0 |
| `react-native-worklets` | 0.12.1 |
| `react-native-gesture-handler` | 3.2.1 |

Platform: native (Android / iOS).

## Steps to reproduce

1. Install the versions listed above.
2. Render a `BottomSheet` containing `BottomSheetScrollView`.
3. Start a native development build.
4. Open the screen containing the bottom sheet.
5. Observe the Reanimated warning.

## Reproduction

```tsx
import React from 'react';
import { Text } from 'react-native';
import BottomSheet, {
  BottomSheetScrollView,

export function Example() {
  return (
    <BottomSheet index={0} snapPoints={['50%']}>
      <BottomSheetScrollView>
        <Text>Content</Text>
      </BottomSheetScrollView>
    </BottomSheet>
  );
}
```

## Actual behavior

The following warning is logged:

```text
[Reanimated] dependencies should only be used in web implementation.

at useHandler (useHandler.native.ts)
at useAnimatedScrollHandler (useAnimatedScrollHandler.ts)
at useScrollHandler (useScrollHandler.ts)
at BottomSheetScrollView (createBottomSheetScrollableComponent.tsx)
```

## Expected behavior

`BottomSheetScrollView` should not pass an explicit dependency array to Reanimated hooks on native platforms.

## Suspected cause

`useScrollHandler` passes a dependency array as the second argument of `useAnimatedScrollHandler`:

https://github.com/gorhom/react-native-bottom-sheet/blob/v5.2.14/src/hooks/useScrollHandler.ts

Reanimated 4.6.0 forwards this argument to its native `useHandler` implementation. The implementation now warns whenever `dependencies !== undefined`:

https://github.com/software-mansion/react-native-reanimated/blob/4.6.0/packages/react-native-reanimated/src/hook/useHandler.native.ts

According to the Reanimated API, dependency arrays are only relevant on Web when the Worklets Babel plugin is unavailable.

## Suggested fix

Preserve the dependency array on Web, but pass `undefined` on native:

```tsx
import { Platform } from 'react-native';

const scrollHandler = useAnimatedScrollHandler(
  handlers,
  Platform.OS === 'web' ? dependencies : undefined
);
```

A platform-specific implementation such as `useScrollHandler.web.ts` and `useScrollHandler.native.ts` would also solve the issue.

This keeps Web-without-plugin dependency tracking while avoiding the native warning introduced by Reanimated 4.6.0.

### Reproduction sample

https://snack.expo.dev/@yoy123/bottomsheetscrollview---reanimated-4_6_0-dependency-warning

### Relevant log output

```shell
[Reanimated] dependencies should only be used in web implementation.
```

huytdps13400:
## Motivation

Fixes #2758.

Reanimated 4.6 warns whenever native `useAnimatedScrollHandler` receives a dependency array because that argument is only intended for web builds without the Worklets Babel plugin. Bottom Sheet already resolves `useScrollHandler.web.ts` on web, so `useScrollHandler.ts` is the native implementation and does not need the second argument.

This removes only the native dependency array. The handler closures and existing platform-specific web implementation remain unchanged.

## Verification

- TDD regression: the TypeScript-AST contract test failed with native argument count `2`, then passed with `1`
- `yarn test` — 1 passed
- `yarn typescript` — passed
- `yarn build` — CommonJS, module, and declarations passed; built native output calls `useAnimatedScrollHandler` with one argument
- Biome check for all changed files — passed

The repository-wide `yarn lint` command still reports one existing warning in unchanged `src/hooks/useBoundingClientRect.ts:54` (`useOptionalChain`); this PR adds no lint findings.

I intentionally did not migrate the existing `runOnJS` calls here: the package still supports Reanimated 3 and does not declare `react-native-worklets` as a peer, so that compatibility migration is separate from this warning-only fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   "homepage": "https://gorhom.dev/react-native-bottom-sheet/",
   "scripts": {
     "typescript": "tsc --skipLibCheck --noEmit",
+    "test": "node --test",
     "lint": "biome lint --error-on-warnings ./src",
     "build": "bob build && yarn copy-dts && yarn delete-dts.js && yarn delete-debug-view",
     "copy-dts": "copyfiles -u 1 \"src/**/*.d.ts\" lib/typescript",
--- a/src/hooks/useScrollHandler.ts
+++ b/src/hooks/useScrollHandler.ts
@@ -30,43 +30,31 @@ export const useScrollHandler = (
   } = useScrollEventsHandlers(scrollableRef, scrollableContentOffsetY);
 
   // callbacks
-  const scrollHandler = useAnimatedScrollHandler(
-    {
-      onScroll: (event, context) => {
-        handleOnScroll(event, context);
+  const scrollHandler = useAnimatedScrollHandler({
+    onScroll: (event, context) => {
+      handleOnScroll(event, context);
 
-        if (onScroll) {
-          runOnJS(onScroll)({ nativeEvent: event });
-        }
-      },
-      onBeginDrag: (event, context) => {
-        handleOnBeginDrag(event, context);
-
-        if (onScrollBeginDrag) {
-          runOnJS(onScrollBeginDrag)({ nativeEvent: event });
-        }
-      },
-      onEndDrag: (event, context) => {
-        handleOnEndDrag(event, context);
-
-        if (onScrollEndDrag) {
-          runOnJS(onScrollEndDrag)({ nativeEvent: event });
-        }
-      },
-      onMomentumBegin: handleOnMomentumBegin,
-      onMomentumEnd: handleOnMomentumEnd,
+      if (onScroll) {
+        runOnJS(onScroll)({ nativeEvent: event });
+      }
     },
-    [
-      handleOnScroll,
-      handleOnBeginDrag,
-      handleOnEndDrag,
-      handleOnMomentumBegin,
-      handleOnMomentumEnd,
-      onScroll,
-      onScrollBeginDrag,
-      onScrollEndDrag,
-    ]
-  );
+    onBeginDrag: (event, context) => {
+      handleOnBeginDrag(event, context);
+
+      if (onScrollBeginDrag) {
+        runOnJS(onScrollBeginDrag)({ nativeEvent: event });
+      }
+    },
+    onEndDrag: (event, context) => {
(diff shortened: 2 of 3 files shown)


jslok:
Same warning also comes from dependency arrays passed to `useDerivedValue` / `useAnimatedReaction` / `useAnimatedStyle` elsewhere in `src/` (5.2.14, Reanimated 4.x, native):

- `components/bottomSheet/BottomSheet.tsx`, `BottomSheetBody.tsx`, `BottomSheetContent.tsx`
- `components/bottomSheetBackdrop/BottomSheetBackdrop.tsx`
- `components/bottomSheetFooter/BottomSheetFooter.tsx`, `BottomSheetFooterContainer.tsx`
- `components/bottomSheetRefreshControl/BottomSheetRefreshControl.android.tsx`
- `components/bottomSheetScrollable/createBottomSheetScrollableComponent.tsx`
- `components/bottomSheetDebugView/ReText.tsx`
- `hooks/useAnimatedDetents.ts`, `useAnimatedKeyboard.ts`, `useAnimatedLayout.ts`, `useBottomSheetContentContainerStyle.ts`, `useScrollable.ts`

Since `warnOnce` dedupes on the message, fixing only `useScrollHandler` still leaves the warning on startup. Unlike `useScrollHandler`, most of these have no `.web` variant, so they'd need the deps gated to web rather than dropped outright.

